PR #74, opened 2021-03-11
feat(hooks): add use-query-params

philibea:
Migration of tiny hooks useQueryParams
What i did : 
- add unit test 
- avoid useless re-render

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ scaleway-lib is a set of NPM packages used at Scaleway.
 - [`@scaleway/countries`](./packages/countries/README.md): ISO 3166/3166-2 coutries JSON database
 - [`@scaleway/eslint-config-react`](./packages/eslint-config-react/README.md): A shared eslint react opiniated configuration
 - [`@scaleway/random-name`](./packages/random-name/README.md): A tiny utility to generate random names
+- [`@scaleway/use-query-params`](./packages/use-query-params/README.md): A tiny react hook to read and update URLs query parameters
 - [`@scaleway/regex`](./packages/regex/README.md): usefull regex named
 
 ## Development
--- a/packages/use-query-params/README.md
+++ b/packages/use-query-params/README.md
@@ -1,0 +1,60 @@
+# `@scaleway/use-query-params`
+
+## A tiny hooks to handle use-query-params
+
+## Install
+
+```bash
+$ yarn add @scaleway/use-query-params
+```
+
+## Usage
+
+```js
+import React from 'react'
+import useQueryParams from '@scaleway/use-query-params'
+
+// this component should be wrap with a Router
+const Component = () => {
+  const { queryParams, setQueryParams } = useQueryParams()
+  const { user } = queryParams
+  const setUser = () => setQueryParams({ user: 'John' })
+  // ?user=John
+
+  return (
+    <>
+      <h1>User: {user}</h1>
+      <button onClick={setUser}>Set User John</button>
+    </>
+  )
+}
+```
+
+## Usage with a custom updater
+
+```js
+import React from 'react'
+import useQueryParams from '@scaleway/use-query-params'
+
+// this component should be wrap with a Router
+const Component = () => {
+  const updater = (prevState, nextState) => ({
+    ...prevState,
+    ...Object.keys(nextState).reduce(
+      (acc, key) => ({ ...acc, [key]: nextState[key].toUpperCase() }),
+      {},
+    ),
+  })
+  const { queryParams, setQueryParams } = useQueryParams({ updater })
+  const { user } = queryParams
+  const setUser = () => setQueryParams({ user: 'John' })
+  // ?user=JOHN
+
+  return (
+    <>
+      <h1>User: {user}</h1>
+      <button onClick={setUser}>Set User John</button>
+    </>
+  )
+}
+```
--- a/packages/use-query-params/package.json
+++ b/packages/use-query-params/package.json
@@ -1,0 +1,45 @@
+{
+  "name": "@scaleway/use-query-params",
+  "version": "1.0.0",
+  "description": "A small hook to handle params",
+  "keywords": [
+    "react",
+    "react-dom",
+    "reactjs",
+    "hooks",
+    "react-router-dom",
+    "params",
+    "query-params"
+  ],
+  "main": "dist/index.js",
+  "module": "dist/module.js",
+  "browser": {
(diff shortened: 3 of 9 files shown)


chambo-e:
You'll also need to update the main README to add the reference to this package :)

packages/use-query-params/src/__tests__/index.js
```diff
@@ -0,0 +1,149 @@
...
+    })
+  })
+
+  it('should not update when location change', () => {
+    const { result } = renderHook(
+      () => useQueryParam({ updateOnChange: false }),
+      {
+        // eslint-disable-next-line react/prop-types
+        wrapper: ({ children }) => (
+          <MemoryRouter
```
chambo-e, line 126: As this Wrapper is the same across multiple tests it would be nice to create a variable in top of the tests :)

packages/use-query-params/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+  const setQueryParams = useCallback(
+    updater => {
+      const currentQueryParams = parse(location.search)
+
+      const newParams =
+        updater instanceof Function
+          ? updater(currentQueryParams)
+          : defaultFnUpdater(currentQueryParams, updater)
+
+      const stringifiedParams = stringify(newParams)
```
chambo-e, line 26: I just noticed, another option instead if using a deep-equal to avoid useless query string updates would be to use a consistent sort in the stringify

For example:

```js
const stringifiedParams = stringify(newParams, { sort: (a, b) => order.indexOf(a) - order.indexOf(b) })

if (stringifiedParams === location.search) {
  return newParams
}
```

packages/use-query-params/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@scaleway/use-query-params",
+  "version": "1.0.0",
+  "description": "A small hooks to handle params",
```
chambo-e, line 4:
```suggestion
  "description": "A small hook to handle params",
```

README.md
```diff
@@ -27,6 +27,7 @@ scaleway-lib is a set of NPM packages used at Scaleway.
 - [`@scaleway/countries`](./packages/countries/README.md): ISO 3166/3166-2 coutries JSON database
 - [`@scaleway/eslint-config-react`](./packages/eslint-config-react/README.md): A shared eslint react opiniated configuration
 - [`@scaleway/random-name`](./packages/random-name/README.md): A tiny utility to generate random names
+- [`@scaleway/use-query-params`](./packages/use-query-params/README.md): A tiny react hooks to handle params
```
chambo-e, line 30:
```suggestion
- [`@scaleway/use-query-params`](./packages/use-query-params/README.md): A tiny react hook to handle URLs query parameters
```

packages/use-query-params/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+      return state
+    },
+    [parseFormat, defaultFnUpdater, search, state, replace, pathname],
+  )
+
+  useEffect(() => {
+    setQueryParams()
+  }, [setQueryParams])
+
+  return {
```
chambo-e, line 53: Shouldn't we return an array here to stay consistent with react hooks ?

packages/use-query-params/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+        sort: (a, b) => a.localeCompare(b),
+      })
+
+      if (stringifiedParams !== search.replace('?', '')) {
+        replace(`${pathname}?${stringifiedParams}`)
+
+        setState(params)
+      }
+
+      return state
```
chambo-e, line 44: Isn't this a bad practice ? If you want to read from the hook you will use the returned hook value no ?

```js
const [queryParams, setQueryParams] = useQueryParam()

// bad
const newQueryParams = setQueryParams({ foo: 'bar' })
console.log(newQueryParams)

// good
setQueryParams({ foo: 'bar' })
console.log(queryParams)
```

packages/use-query-params/src/index.js
chambo-e: Nevermind, would only be useful if we would use this hook multiple times in the same component

packages/use-query-params/src/index.js
philibea: Yes you totally right !

packages/use-query-params/src/index.js
philibea: I believe it's a good practice if the hook will be use more than one time in a component 

```javascript
So that you can more easily do ...

const [count1, setCount1] = useState(0);
const [count2, setCount2] = useState(0);

With an object you'd need to do ...

const { count: count1, setCount: setCount1 } = useState(0);
const { count: count2, setCount: setCount2 } = useState(0);
```

packages/use-query-params/README.md
```diff
@@ -0,0 +1,34 @@
...
+## Install
+
+```bash
+$ yarn add @scaleway/use-query-params
+```
+
+## Usage
+
+```js
+import React from 'react'
```
chambo-e, line 14: Might be good to add an example with a custom updater

philibea: Done ;)

packages/use-query-params/README.md
```diff
@@ -0,0 +1,58 @@
...
+import useQueryParams from '@scaleway/use-query-params'
+
+// this component should be wrap with a Router
+const Component = () => {
+    const updater = (prevState, nextState) => nextState
+    const { getQueryParams, setQueryParams } = useQueryParams({ updater })
+    const { user } = getQueryParams
+    const setUser = () => setQueryParams({ user: "John" })
```
chambo-e, line 46:
```suggestion
    const setUser = () => setQueryParams({ user: "John" })
    // ?user=JOHN
```

packages/use-query-params/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Usage with a custom updater
+
+```js
+import React from 'react'
+import useQueryParams from '@scaleway/use-query-params'
+
+// this component should be wrap with a Router
+const Component = () => {
+    const updater = (prevState, nextState) => nextState
```
chambo-e, line 43:
```suggestion
    const updater = (prevState, nextState) => ({
      ...prevState,
      ...Object.keys(nextState).reduce((acc, key) => ({ ...acc, [key]: nextState[key].toUpperCase()  }))
    })
```

chambo-e (approved):

matthprost (approved):

PR #81, opened 2021-03-17
fix(query-params):re-render hooks

philibea:


packages/use-query-params/src/index.js
```diff
@@ -1,62 +1,32 @@
...
+  }, [])
 
   useEffect(() => {
     const stringifiedParams = stringyFormat(state)
-    if (stringifiedParams !== search.replace('?', '')) {
-      replace(`${pathname}?${stringifiedParams}`)
-    }
-  }, [pathname, replace, search, state, stringyFormat])
-
+    window.history.replaceState({}, null, `${pathname}?${stringifiedParams}`)
```
chambo-e, line 28:
```suggestion
    window.history.replaceState(window.history.state, null, `${pathname}?${stringifiedParams}`)
```

chambo-e (changes requested):
Also need to remove react-router-dom dependency

chambo-e (approved):

DorianMaliszewski (approved):

PR #85, opened 2021-03-19
fix: query params hook

DorianMaliszewski:
## Types

- Fix

## What I did

- Usage of useEffect with cleanUp and debounce like to fix update loop


matthprost (approved):

philibea (approved):

chambo-e:
I quite don't understand this PR, we return to the old usage with react-router. ?

PR #94, opened 2021-03-24
fix(use-query-params): correct infinite loop when search is empty

chambo-e:
When initial search is empty we have an infinite loop

As search is empty and our state is also empty but we compare 

```js
search !== `?${stringifiedParams}`
```

We call replace every 500ms for nothing


philibea (approved):

DorianMaliszewski (approved):
🚀

PR #96, opened 2021-03-25
feat!: base hook on search not own state

DorianMaliszewski:
Why did not just use Router Context ?
- It cant be tested properly 😅 

By adding a context on top of this hook we make possible to useQueryParams everywhere


chambo-e (approved):
🔥

philibea (approved):

PR #130, opened 2021-04-15
feat: add hexa regex

vincentaudebert:
Adds hexadecimal regex to regex lib so it can be used for validators


chambo-e (approved):

chambo-e (approved):
A `feat` would be better ?

matthprost (approved):

PR #134, opened 2021-04-20
feat: use dataloader

DorianMaliszewski:


packages/use-dataloader/README.md
```diff
@@ -0,0 +1,83 @@
+# `@scaleway/use-query-params`
```
philibea, line 1: documentation about use-dataloader and not use-query-params ahah

packages/use-dataloader/src/usePaginatedDataLoader/index.js
```diff
@@ -0,0 +1,219 @@
+import React from 'react'
```
philibea, line 1: Do we have to import React ? it's not jsx, it's a possible gain on bundle size

packages/use-dataloader/src/useDataLoader/index.js
```diff
@@ -0,0 +1,145 @@
...
+  return {
+    state: status,
+    isLoading,
+    isIdle,
+    isSuccess,
+    isError,
+    isPolling,
+    data,
+    error,
+    updateData,
```
chambo-e, line 140: This means the user will be able to manually update the result of a cached query from outside the data loader scope ? 
Shouldn't this function be kept private ?

packages/use-dataloader/src/DataLoaderProvider/index.js
```diff
@@ -0,0 +1,148 @@
...
+    if (key) {
+      return reloads.current[key] || undefined
+    }
+
+    return reloads.current
+  }, [])
+
+  const value = useMemo(
+    () => ({
+      addCachedData,
```
chambo-e, line 116: If I understand correctly most of this context should only be used internally ? 

I also don't really understand the value of `reload` vs `getReloads`, these functions basically have the same result no ?

chambo-e (changes requested):

packages/use-dataloader/README.md
```diff
@@ -0,0 +1,133 @@
...
+  }
+}
+
+export default MyComponent
+```
+
+### useDataLoaderContext
+
+```js
+import { usePaginatedDataLoader } from '@scaleway-lib/use-dataloader'
```
philibea, line 78: Not use anymore ?

packages/use-dataloader/src/useDataLoader/index.js
DorianMaliszewski: The user can change data manually if he want to edit them (to use it in a form or something else)

packages/use-dataloader/src/DataLoaderProvider/index.js
DorianMaliszewski: This context handle data cache and all function associated with their key. Like this you can trigger data reload from another component.  `reload` directly trigger a reload where as `getReloads` is just to retrieve a reload function so by exemple: 
```js
const dataloaderContext = useDataLoaderContex()
const reloadOtherComponent = () => {
  dataloaderContext.reload("test-key") // Reload only test-key function
  dataloaderContext.reload() // Reload all functions
}
// Not as easy as
const reloadOtherComponent = () => {
  dataloaderContext.getReloads("test-key")?.() // Reload test-key function
  dataloaderContext.getReloads().forEach(fn => fn()) // Reload all functions
}
```

But you can have a specific case where you need to get only the reload function, To ensure it exist or anything else

packages/use-dataloader/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "url": "https://github.com/scaleway/scaleway-lib",
+    "directory": "packages/use-dataloader"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "react": "17.x",
+    "react-dom": "17.x"
```
chambo-e, line 34: We have no implicit react-dom dependency right (i.e. we could use this package with react-native) ?

packages/use-dataloader/src/useDataLoader.js
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }, [keepOldData, data])
+
+  return {
+    isLoading,
+    isIdle,
+    isSuccess,
+    isError,
+    isPolling,
+    oldData: oldDataRef.current,
```
chambo-e, line 155: `previousData`, old is time related :)

packages/use-dataloader/src/reducer.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable consistent-return */
```
chambo-e, line 1: Shouldn't we have a default return ? Maybe returning the current state to avoid undefined behaviour ? Would also allow us to remove the eslint silencing :)

chambo-e (changes requested):
Some last cosmetic feedbacks 🖌️ 

On a personal level I'm not a fan of exposing the `updateData` function in the hook, I don't think this is the library concern to provide it. An user might wonder why the data returned from its fetch call is different from the one he receive from the hook.

I'm also kinda unsure about exposing that widely fonctions from the context, it might lead to bad practices to be able to retrieve/update reload functions and cache from outside the hook scope.

But I don't mind going this way and reevaluating in a couple weeks/months 🙂 

Overall you've done a massive job, congrats 🔥

packages/use-dataloader/package.json
```diff
@@ -0,0 +1,43 @@
...
+  },
+  "devDependencies": {
+    "@testing-library/jest-dom": "^5.11.9",
+    "@testing-library/react": "^11.2.5",
+    "@testing-library/react-hooks": "^5.1.1",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1"
```
philibea, line 41: We didn't use it with our tests so it's not necessary

packages/use-dataloader/src/useDataLoader.js
```diff
@@ -0,0 +1,164 @@
...
+    clearReload,
+    getCachedData,
+  } = useDataLoaderContext()
+  const [{ status, data, error }, dispatch] = useReducer(reducer, {
+    status: StatusEnum.IDLE,
+    data: initialData,
+    error: undefined,
+  })
+
+  const oldDataRef = useRef()
```
philibea, line 46: previousDataRef = useRef()

 old is time related :)

packages/use-dataloader/src/useDataLoader.js
```diff
@@ -0,0 +1,164 @@
...
+    reloadOnKeyChange,
+    initialData,
+  ])
+
+  useLayoutEffect(() => {
+    methodRef.current = method
+  }, [method])
+
+  useLayoutEffect(() => {
+    if (keepOldData && data && oldDataRef.current !== data) {
```
philibea, line 144: if (keepPreviousData && data && previousDataRef.current !== data) {

packages/use-dataloader/README.md
```diff
@@ -0,0 +1,133 @@
...
+    clearCachedData('test') // Clear only the cached data of the test key
+    clearCachedData() // Clear all cached data
+  }
+
+  useEffect(() => {
+    addReloads('test', fakePromise); // Add a reload in the context
+
+    return () => {
+      clearReloads('test') // Clear only the reload with the test key
+      clearReloads() // Clear all reloads
```
philibea, line 118: if you forget the key, it's a little bit dangerous to have reloads that will reload all your data in your applications ?

packages/use-dataloader/src/DataLoaderProvider.js
```diff
@@ -0,0 +1,148 @@
...
+          return tmp
+        })
+      } else {
+        setCachedData({})
+      }
+    },
+    [setCachedData],
+  )
+
+  const reload = useCallback(async key => {
```
philibea, line 88: Reload should be separate in two:
- one only scope to this dataLoader
- one can be used to reload all app ( maybe reloadAllDataLoader ? )

packages/use-dataloader/src/DataLoaderProvider.js
```diff
@@ -0,0 +1,148 @@
...
+
+  const setReloads = useCallback(compute => {
+    if (typeof compute === 'function') {
+      reloads.current = compute(reloads.current)
+    } else {
+      reloads.current = compute
+    }
+  }, [])
+
+  const addCachedData = useCallback(
```
philibea, line 32: addCachedData should be implicite and use everytime we make a load to keep fresh data in the cache.

packages/use-dataloader/src/__tests__/useDataLoader.test.js
```diff
@@ -0,0 +1,450 @@
...
+      props => useDataLoader(props.key, props.method, props.config),
+      {
+        wrapper,
+        initialProps,
+      },
+    )
+    expect(result.current.data).toBe(undefined)
+    expect(result.current.isLoading).toBe(true)
+    rerender()
+    rerender()
```
philibea, line 34: do we need two rerender() here ?

packages/use-dataloader/src/useDataLoader.js
```diff
@@ -0,0 +1,164 @@
...
+  const isLoading = useMemo(() => status === StatusEnum.LOADING, [status])
+  const isIdle = useMemo(() => status === StatusEnum.IDLE, [status])
+  const isSuccess = useMemo(() => status === StatusEnum.SUCCESS, [status])
+  const isError = useMemo(() => status === StatusEnum.ERROR, [status])
+  const isPolling = useMemo(
+    () => enabled && pollingInterval && (isSuccess || isLoading),
+    [isSuccess, isLoading, enabled, pollingInterval],
+  )
+
+  const updateData = useCallback(
```
philibea, line 61: udpateDate should not be used outside of the dataLoader, data is always true from api and not from our side if we can manipulate it

packages/use-dataloader/src/useDataLoader.js
```diff
@@ -0,0 +1,164 @@
...
+  key,
+  method,
+  {
+    onSuccess,
+    onError,
+    initialData,
+    pollingInterval,
+    enabled = true,
+    reloadOnKeyChange = true,
+    keepOldData = true,
```
philibea, line 31: keepPreviousData or more implicite keepCachedData but this should be always true if we have key or cacheKey ?

packages/use-dataloader/package.json
DorianMaliszewski: I think yes I dropped it

packages/use-dataloader/src/reducer.js
DorianMaliszewski: No we can't. If we do that we can't have 100% test coverage (without importing the reducer himself)

chambo-e (approved):

packages/use-dataloader/src/useDataLoader/index.js
philibea: For me it's a bad practice to let the user transform de data.

Imagine that you modify the date before a reload and you make put/patch/post request that goes wrong ? You will need to clear the data in this case ?
if you forget to clear the data you will show wrong information to your users ?

An other point :
you believe that the api will return everything you put in updateData but if there an error again and the data doesn't return what you think she will return. it's will not break at first but if you comme back later on this page and you didn't handle the use case where for exemple de data is a number and not a string ? or a boolean and not a number this will break your app

philibea (approved):

PR #145, opened 2021-04-29
fix(use-i18n)!: move date-fns dynamic import outside of library scope

chambo-e:
This was causing a SIGSEGV for some reason on our tests

BREAKING CHANGE: We now expose loadDateLocale props in the Provider instead of bundling it

### DIFF
--- a/packages/use-i18n/README.md
+++ b/packages/use-i18n/README.md
@@ -27,11 +27,11 @@ Use of local `variables` and `namespace` to dynamically load locales.
  ┃ ┗ 📜common.json
 ```
 
-your loader will be:
+your loaders will be:
 
 ```js
-const load = ({ locale, namespace }) =>
-  import(`./locales/${locale}/${namespace}`)
+const load = ({ locale, namespace }) => import(`./locales/${locale}/${namespace}`)
+const loadDateLocale = (locale) => import(`date-fns/locale/${locale}/index`)
 ```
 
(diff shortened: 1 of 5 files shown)


packages/use-i18n/README.md
```diff
@@ -27,11 +27,11 @@ Use of local `variables` and `namespace` to dynamically load locales.
  ┃ ┗ 📜common.json
 ```
 
-your loader will be:
+your loaders will be:
 
 ```js
-const load = ({ locale, namespace }) =>
-  import(`./locales/${locale}/${namespace}`)
+const load = ({ locale, namespace }) => import(`./locales/${locale}/${namespace}`)
```
philibea, line 33: const load = async ({ locale, namespace }) => import(`./locales/${locale}/${namespace}`)

packages/use-i18n/README.md
```diff
@@ -27,11 +27,11 @@ Use of local `variables` and `namespace` to dynamically load locales.
...
+const loadDateLocale = (locale) => import(`date-fns/locale/${locale}/index`)
```
philibea, line 34: const loadDateLocale = async (locale) => import(`date-fns/locale/${locale}/index`)

philibea (changes requested):

matthprost (approved):

philibea (approved):

PR #183, opened 2021-05-17
feat(i18n): add loaded namespaces and isLoaded into useTranslation 

philibea:
Add a new state namespaces on Provider, this state re-present namespace that are already loaded.
Add isLoaded on useTranslation hook, this params will help to avoid multi-rerender when it's not needed.
Add ReactDom unstable_batchedUpdates  to minimize number of renders

### DIFF
--- a/packages/use-i18n/package.json
+++ b/packages/use-i18n/package.json
@@ -34,11 +34,13 @@
   },
   "peerDependencies": {
     "date-fns": "2.x",
-    "react": "17.x"
+    "react": "17.x",
+    "react-dom": "17.x"
   },
   "devDependencies": {
     "@testing-library/react-hooks": "^6.0.0",
     "mockdate": "^3.0.5",
-    "react": "17.0.2"
+    "react": "17.0.2",
+    "react-dom": "17.0.2"
   }
 }
--- a/packages/use-i18n/src/index.js
+++ b/packages/use-i18n/src/index.js
@@ -10,6 +10,7 @@ import React, {
   useMemo,
   useState,
 } from 'react'
+import ReactDOM from 'react-dom'
 import 'intl-pluralrules'
 
 const LOCALE_ITEM_STORAGE = 'locale'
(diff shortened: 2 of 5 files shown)


packages/use-i18n/package.json
```diff
@@ -30,15 +30,18 @@
     "intl-format-cache": "^4.3.1",
     "intl-messageformat": "^9.5.3",
     "intl-pluralrules": "^1.2.2",
-    "prop-types": "^15.7.2"
+    "prop-types": "^15.7.2",
+    "react-dom": "^17.0.2"
```
chambo-e, line 34: Not needed :)

packages/use-i18n/src/index.js
```diff
@@ -235,15 +250,15 @@ const I18nContextProvider = ({
...
+      formatNumber,
       loadTranslations,
-      locales,
       namespaceTranslation,
+      namespaces,
       relativeTime,
       relativeTimeStrict,
-      setLocales,
       setTranslations,
+      supportedLocales,
```
chambo-e, line 261: What the difference between supportedLocales and locales ?

packages/use-i18n/src/index.js
```diff
@@ -114,15 +130,22 @@ const I18nContextProvider = ({
...
       const preparedValues = prefix ? prefixKeys(`${prefix}.`)(values) : values
 
-      setTranslations(prevState => ({
-        ...prevState,
-        ...{
-          [currentLocale]: {
-            ...prevState[currentLocale],
-            ...preparedValues,
+      // avoid a lot of render when async update
+      ReactDOM.unstable_batchedUpdates(() => {
```
vincentaudebert, line 134: any idea why is it "unstable"?

packages/use-i18n/src/index.js
philibea: Before we use a state and it's not used anymore

packages/use-i18n/src/index.js
vincentaudebert: Ok got it in your article

Although this function is supposedly “unstable”, React apparently intends to address this in the following versions.

    “In future versions (probably React 17 and later), React will batch all updates by default so you won’t have to think about this”, according to Dan Abramov.

vincentaudebert (approved):

packages/use-i18n/src/index.js
philibea: It's unstable for the moment but it's use in React, Redux library from a long time. 
In a future upgrade of react, i thinks it's will not necessary

chambo-e (approved):

PR #196, opened 2021-05-25
feat: use dataloader handle enabled changes

DorianMaliszewski:
# What I did

- Optimization of hooks rerender by using some refs
- Handle enabled change and key change better

### DIFF
--- a/packages/use-dataloader/src/useDataLoader.js
+++ b/packages/use-dataloader/src/useDataLoader.js
@@ -51,15 +51,17 @@ const useDataLoader = (
     pollingInterval,
   } = {},
 ) => {
-  const { addCachedData, addReload, clearReload, getCachedData } =
+  const { addReload, clearReload, getCachedData, addCachedData } =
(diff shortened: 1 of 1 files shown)


packages/use-dataloader/src/useDataLoader.js
```diff
@@ -51,15 +51,16 @@ const useDataLoader = (
     pollingInterval,
   } = {},
 ) => {
-  const { addCachedData, addReload, clearReload, getCachedData } =
-    useDataLoaderContext()
+  const dataLoaderContext = useDataLoaderContext()
```
matthprost, line 54: I think you can destructure both functions you are using from `useDataLoaderContext()`:

```js
  const { addCachedData, getCachedData } = useDataLoaderContext()
```

matthprost (changes requested):

matthprost (approved):

chambo-e (approved):
Looks good :)

PR #256, opened 2021-07-02
feat: return result from promise

DorianMaliszewski:


chambo-e:
👋 I don't really understand why, do you have an example use case ?

DorianMaliszewski:
> 👋 I don't really understand why, do you have an example use case ?

@chambo-e If you need to get the response inside a function directly from the `reload` method

chambo-e:
> @chambo-e If you need to get the response inside a function directly from the `reload` method

This is quite weird no ? This means we can access the data in two different ways ?

PR #325, opened 2021-08-11
feat(use-i18n): allow to specify a prefix in the hook

chambo-e:


philibea (approved):

packages/use-i18n/src/usei18n.tsx
```diff
@@ -84,13 +84,16 @@ interface Context {
...
+export const useI18n = (prefix?: string): Context => {
   const context = useContext(I18nContext)
   if (context === undefined) {
     throw new Error('useI18n must be used within a I18nProvider')
   }
 
-  return context
+  return {
+    ...context,
+    t: prefix ? context.namespaceTranslation?.(prefix) : context.t,
```
DorianMaliszewski, line 95: Doesn't it be better to fallback to the original `t` in case if the translation is not found with the prefix ?

chambo-e: That's what the namespaceTranslation does :)

DorianMaliszewski: DIdn't see that translate was the default 2nd param 😅

DorianMaliszewski (approved):

PR #355, opened 2021-08-31
feat(use-query-params): add ability to push instead of replacing current history

chambo-e:


philibea (approved):

packages/use-query-params/README.md
```diff
@@ -86,3 +86,28 @@ const Component = () => {
...
+
+To avoid mutating history
+
+```js
+// In this exemple we assume that we have an URL that include : `?company=Scaleway".
+import React from 'react'
+import useQueryParams from '@scaleway/use-query-params'
+
+const Component = () => {
+  const { queryParams, setQueryParams } = useQueryParams()
```
adriengibrat, line 100:
```suggestion
  const { queryParams, replaceQueryParams } = useQueryParams()
```

chambo-e: We can use both :)

adriengibrat (approved):

packages/use-query-params/README.md
adriengibrat: example use `replaceQueryParams` just below

chambo-e: Right, sorry my bad :)

PR #369, opened 2021-09-08
feat: request cancellation

DorianMaliszewski:
Possibility to cancel a Promise with an enhanced promise type

### DIFF
--- a/packages/use-dataloader/src/__tests__/useDataLoader.tsx
+++ b/packages/use-dataloader/src/__tests__/useDataLoader.tsx
@@ -1,7 +1,20 @@
+/* eslint-disable no-console */
 import { act, renderHook } from '@testing-library/react-hooks'
 import React from 'react'
 import DataLoaderProvider, { useDataLoaderContext } from '../DataLoaderProvider'
-import useDataLoader from '../useDataLoader'
+import useDataLoader, {
+  PromiseType,
+  UseDataLoaderConfig,
+  UseDataLoaderResult,
+} from '../useDataLoader'
+
+type UseDataLoaderHookProps = {
+  config: UseDataLoaderConfig<unknown>
+  key: string
+  method: () => Promise<unknown>
+}
+
+const PROMISE_TIMEOUT = 100
 
 const initialProps = {
   config: {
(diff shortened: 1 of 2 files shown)


packages/use-dataloader/src/__tests__/useDataLoader.tsx
```diff
@@ -15,27 +25,29 @@ const initialProps = {
     }),
 }
 // eslint-disable-next-line react/prop-types
-const wrapper = ({ children }: { children: React.ReactNode }) => (
+const wrapper = ({ children }: { children?: React.ReactNode }) => (
```
vincentaudebert, line 28: Le declarer en VoidFunctionComponent?

vincentaudebert:
I get this 

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
          at TestComponent (/Volumes/Projets/scaleway-lib/node_modules/@testing-library/react-hooks/lib/helpers/createTestHarness.js:22:5)
          at DataLoaderProvider (/Volumes/Projets/scaleway-lib/packages/use-dataloader/src/DataLoaderProvider.tsx:33:31)
          at wrapper (/Volumes/Projets/scaleway-lib/packages/use-dataloader/src/__tests__/useDataLoader.tsx:28:20)
          at Suspense
          at ErrorBoundary (/Volumes/Projets/scaleway-lib/node_modules/react-error-boundary/dist/react-error-boundary.cjs.js:40:35)

      123 |     async cacheKey => {
      124 |       try {
    > 125 |         dispatch(Actions.createOnLoading())
          |         ^
      126 |         const promise = method()
      127 |         cancelMethodRef.current = promise.cancel
      128 |         const result = await promise.then(res => res)
```

in the tests. Is there no way to avoid this?

packages/use-dataloader/src/__tests__/useDataLoader.tsx
```diff
@@ -582,4 +608,73 @@ describe('useDataLoader', () => {
...
+    const { result, unmount } = renderHook<
+      UseDataLoaderHookProps,
+      UseDataLoaderResult
+    >(
+      props =>
+        useDataLoader(
+          props.key,
+          () => {
+            const promise = new Promise(resolve => {
+              setTimeout(() => resolve(true), 500)
```
adriengibrat, line 623: I know it already was the value used for other tests, but 500ms seems too long IMHO.

adriengibrat (dismissed):

adriengibrat (commented):
Why not using the native fetch abort mechanism?
> the fetch() promise rejects with an "AbortError" DOMException.

https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal

packages/use-dataloader/src/__tests__/useDataLoader.tsx
DorianMaliszewski: Changed by 100ms that's better yes

DorianMaliszewski:
> I'm not fan of `enhanced promise type` :
> 
> Why not using the native fetch abort mechanism?
> 
> > the fetch() promise rejects with an "AbortError" DOMException.
> 
> https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal

We should handle both Axios and Fetch behaviour so we need to keep something flexible to ensure a compatibility between both. May in the fututre we will directly integrate a custom hook for each ones.

adriengibrat (approved):

vincentaudebert (approved):
💯

chambo-e (approved):

PR #392, opened 2021-09-21
feat: concurrent request handle with dataloader

DorianMaliszewski:
Handle when there multiple same request to avoid multiple same promise resolution.


packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -7,87 +7,153 @@ import React, {
...
+          ...current,
+          [computeKey(key)]: newRequest,
         }))
+
+        addReload(key, newRequest.launch)
+
+        return newRequest
       }
+      throw new Error('Key should be a string')
     },
```
matthprost, line 136:
```suggestion
      if (!key || typeof key !== 'string') throw new Error('Key should exist and be a string')
        const notifyChanges = (updatedRequest: DataLoader) => {
          setRequestQueue(current => ({
            ...current,
            [computeKey(updatedRequest.key)]: updatedRequest,
          }))
        }
        const newRequest = new DataLoader({ ...args, notify: notifyChanges })
        newRequest.addOnSuccessListener(result => {
          if (result !== undefined && result !== null)
            addCachedData(key, result)
        })
        setRequestQueue(current => ({
          ...current,
          [computeKey(key)]: newRequest,
        }))

        addReload(key, newRequest.launch)

        return newRequest
      }
```

packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -99,70 +165,84 @@ const DataLoaderProvider = ({ children, cacheKeyPrefix, onError }: {
...
-  }, [])
+  const reload = useCallback(
+    async (key?: string) => {
+      if (key && typeof key === 'string') {
+        await (reloads.current[computeKey(key)] &&
+          reloads.current[computeKey(key)]())
+      }
+    },
+    [computeKey],
+  )
```
matthprost, line 188:
```suggestion
  const reload = useCallback(
    async (key?: string) => {
      if (!key || typeof key !== 'string') throw new Error('Key should exist and be a string')
      await (reloads.current[computeKey(key)] && reloads.current[computeKey(key)]())
    },
    [computeKey],
  )
```

packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -99,70 +165,84 @@ const DataLoaderProvider = ({ children, cacheKeyPrefix, onError }: {
...
 
-  const getReloads = useCallback((key?: string) => {
-    if (key) {
-      return reloads.current[key] || undefined
-    }
+  const getReloads = useCallback(
+    (key?: string) => {
+      if (key) {
+        return reloads.current[computeKey(key)] || undefined
+      }
```
matthprost, line 211: To put a throw error on every function if key is not string or it doesn't exists

matthprost (changes requested):

packages/use-dataloader/src/dataloader.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+    } catch (err) {
+      this.status = StatusEnum.ERROR
+      this.error = err as Error
+      this.notify(this)
+      if (this.errorListeners?.length) {
+        this.errorListeners.forEach(
+          listener => void listener?.(this.error as Error),
+        )
+      }
```
matthprost, line 82:
```suggestion
      if (this.errorListeners?.length > 0) {
        this.errorListeners.forEach(
          listener => void listener?.(this.error as Error),
        )
      }
```

packages/use-dataloader/src/dataloader.ts
```diff
@@ -0,0 +1,145 @@
...
+        this.pollingInterval,
+      ) as unknown as number
+    }
+  }
+
+  public cancel = (): void => {
+    this.cancelMethod?.()
+    if (this.cancelListeners.length) {
+      this.cancelListeners.forEach(listener => void listener?.())
+    }
```
matthprost, line 96:
```suggestion
    if (this.cancelListeners.length > 0) {
      this.cancelListeners.forEach(listener => void listener?.())
    }
```

matthprost (changes requested):

matthprost (changes requested):

packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -99,70 +170,84 @@ const DataLoaderProvider = ({ children, cacheKeyPrefix, onError }: {
...
-  const reload = useCallback(async (key?: string) => {
-    if (key && typeof key === 'string') {
-      await (reloads.current[key] && reloads.current[key]())
-    }
-  }, [])
+  const reload = useCallback(
+    async (key?: string) => {
+      if (key && typeof key === 'string') {
+        await (reloads.current[computeKey(key)] &&
+          reloads.current[computeKey(key)]())
```
matthprost, line 189:
```suggestion
        await (reloads.current[computeKey(key)]?.())
```

packages/use-dataloader/src/__tests__/useDataLoader.test.tsx
```diff
@@ -60,25 +58,19 @@ describe('useDataLoader', () => {
...
-      wrapper,
-    })
-    expect(result.current.data).toBe(undefined)
-    expect(result.current.isLoading).toBe(true)
-    await waitForNextUpdate()
-    expect(result.current.data).toBe(undefined)
-    expect(result.current.isSuccess).toBe(true)
-    expect(result.current.isLoading).toBe(false)
+    )
+    expect(result.error?.message).toBe('Key should be a string')
```
matthprost, line 73: Here you can probably import `KEY_IS_NOT_STRING_ERROR` constant and check trough it

matthprost (changes requested):

matthprost (approved):

packages/use-dataloader/src/constants.ts
```diff
@@ -1,4 +1,9 @@
-export const StatusEnum = {
+import { DataLoaderStatus } from './types'
+
+export const StatusEnum: Record<
+  'ERROR' | 'IDLE' | 'LOADING' | 'SUCCESS',
+  DataLoaderStatus
+> = {
```
adriengibrat, line 6: Why not using actal enum ;)
```suggestion
export enum StatusEnum {
```

then replace all `DataLoaderStatus` with the enum type

adriengibrat (changes requested):

packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -7,87 +7,158 @@ import React, {
...
       if (key && typeof key === 'string') {
         setReloads(actualReloads => {
           const tmp = actualReloads
-          delete tmp[key]
+          delete tmp[computeKey(key)]
 
           return tmp
         })
-      }
+      } else throw new Error(KEY_IS_NOT_STRING_ERROR)
```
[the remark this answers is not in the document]
DorianMaliszewski: No like this we ensure user use this lib according to our expectation. This allow us to omit string verification or others and focus on quality

packages/use-dataloader/src/__tests__/DataLoaderProvider.test.tsx
```diff
@@ -0,0 +1,293 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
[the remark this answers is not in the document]
DorianMaliszewski: Because i tried to pass some integers to make sure errors are throw and ensure all is working

packages/use-dataloader/src/dataloader.ts
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable no-void */
```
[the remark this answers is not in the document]
DorianMaliszewski: I can't launch the void promise like we did in useDataloader before if I don't put this 😢

vincentaudebert (approved):

PR #399, opened 2021-09-27
feat: refresh data on max lifetime reached

DorianMaliszewski:
Consider data as outdated when maxLifetime is reached. Allow stalled data detection when there are multiple useDataLoader with same key. (Performance)

### DIFF
--- a/packages/use-dataloader/README.md
+++ b/packages/use-dataloader/README.md
@@ -77,10 +77,10 @@ const failingPromise = async () => {
 }
 
 const App = () => {
-  useDataLoader('local-error',  failingPromise,  {
-    onError: (error) => {
+  useDataLoader('local-error', failingPromise, {
+    onError: error => {
       console.log(`local onError: ${error}`)
-    }
+    },
   })
 
   useDataLoader('error', failingPromise)
@@ -88,7 +88,7 @@ const App = () => {
   return null
 }
 
-const globalOnError = (error) => {
+const globalOnError = error => {
   console.log(`global onError: ${error}`)
 }
 
@@ -219,6 +219,7 @@ const useDataLoader = (
     pollingInterval, // Relaunch the request after the last success
     enabled = true, // Launch request automatically
     keepPreviousData = true, // Do we need to keep the previous data after reload
+    maxDataLifetime, // Max time before previous success data is outdated (in millisecond)
   } = {},
 )
 ```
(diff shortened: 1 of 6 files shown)


packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -24,6 +24,7 @@ type UseDataLoaderInitializerArgs<T = unknown> = {
   method: () => PromiseType<T>
   pollingInterval?: number
   keepPreviousData?: boolean
+  maxDataLifetime?: number
```
matthprost, line 27: Can we add it into documentation with a precision that this value is in milliseconds ?

matthprost (changes requested):

packages/use-dataloader/README.md
```diff
@@ -219,6 +219,7 @@ const useDataLoader = (
     pollingInterval, // Relaunch the request after the last success
     enabled = true, // Launch request automatically
     keepPreviousData = true, // Do we need to keep the previous data after reload
+    maxDataLifetime, // Max time before preivous success data is outdated
```
matthprost, line 222:
```suggestion
    maxDataLifetime, // Max time before previous success data is outdated (in milliseconds)
```

packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -24,6 +24,10 @@ type UseDataLoaderInitializerArgs<T = unknown> = {
   method: () => PromiseType<T>
   pollingInterval?: number
   keepPreviousData?: boolean
+  /**
+   * Max time before data from previous success is considered as outdated
```
matthprost, line 28:
```suggestion
   * Max time before data from previous success is considered as outdated (in milliseconds)
```

packages/use-dataloader/src/types.ts
```diff
@@ -24,6 +24,10 @@ export interface UseDataLoaderConfig<T = unknown> {
   onError?: OnErrorFn
   onSuccess?: OnSuccessFn
   pollingInterval?: number
+  /**
+   * Max time before data from previous success is considered as outdated
```
matthprost, line 28:
```suggestion
   * Max time before data from previous success is considered as outdated (in milliseconds)
```

matthprost (changes requested):

adriengibrat (approved):

chambo-e (approved):

matthprost (approved):

PR #413, opened 2021-10-04
feat: max concurrent request usedataloader

DorianMaliszewski:
Allow user to set max concurrent dataloader in parallel


packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -30,209 +33,186 @@ type UseDataLoaderInitializerArgs<T = unknown> = {
...
 
 // @ts-expect-error we force the context to undefined, should be corrected with default values
-export const DataLoaderContext = createContext<Context>(undefined)
+export const DataLoaderContext = createContext<IDataLoaderContext>(undefined)
 
 const DataLoaderProvider = ({
   children,
   cacheKeyPrefix,
   onError,
+  maxConcurrentRequests,
```
adriengibrat, line 67:
```suggestion
  maxConcurrentRequests = DEFAULT_MAX_CONCURRENT_REQUESTS,
```
plutôt que les `defaultProps`

packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -30,209 +33,186 @@ type UseDataLoaderInitializerArgs<T = unknown> = {
...
             ...current,
-            [computeKey(updatedRequest.key)]: updatedRequest,
+            [updatedRequest.key]: updatedRequest,
           }))
         }
-        const newRequest = new DataLoader({ ...args, notify: notifyChanges })
-        newRequest.addOnSuccessListener(result => {
-          if (result !== undefined && result !== null)
-            addCachedData(key, result)
+        const newRequest = new DataLoader({
```
adriengibrat, line 90: `maxConcurrentRequests`

should be passed here rather than set later with `useLayoutEffect`.

packages/use-dataloader/src/dataloader.ts
```diff
@@ -52,6 +58,17 @@ class DataLoader<T = unknown> {
     this.keepPreviousData = args?.keepPreviousData
     this.notify = args.notify
     this.maxDataLifetime = args.maxDataLifetime
+    if (args.enabled) {
+      const tryLaunch = () => {
+        if (DataLoader.started < DataLoader.maxConcurrent) {
+          // eslint-disable-next-line no-void
+          void this.load()
```
adriengibrat, line 65: I don't get why `void` is for here?

packages/use-dataloader/src/dataloader.ts
```diff
@@ -12,6 +12,10 @@ export type DataLoaderConstructorArgs<T = unknown> = {
 }
 
 class DataLoader<T = unknown> {
+  public static started = 0
+
+  public static maxConcurrent = DEFAULT_MAX_CONCURRENT_REQUESTS
```
adriengibrat, line 17: Is there a reason for `started` & `maxConcurent` to be static?

This could lead to weird side effects, IMO

packages/use-dataloader/src/dataloader.ts
DorianMaliszewski: load is async and promise shouldn't be launched without then/await according to our rules so by putting void it says that this is wanted

packages/use-dataloader/src/dataloader.ts
DorianMaliszewski: Yeah this properties are static like this all instances of DataLoader can share the reference to the value and be all up-to-date. The solid principle is kept because of DataLoaderProvider should only manage dataloader instance and not the request queue. Moreover keeping our complexity of request launches in the class allow us to keep a simple hook and Provider.

packages/use-dataloader/src/DataLoaderProvider.tsx
DorianMaliszewski: useLayoutEffect is triggered before the component will be mount in the DOM so it will always set the maxConcurrentRequests before any `useDataLoader` are mounted.

packages/use-dataloader/src/DataLoaderProvider.tsx
DorianMaliszewski: Because of DataLoader class and Provider have a default value this will never happen

packages/use-dataloader/src/DataLoaderProvider.tsx
adriengibrat: I know when it will be triggered ;) it's just weird to set it elsewhere.

packages/use-dataloader/src/dataloader.ts
adriengibrat: not obvious, I'd prefer an explicit comment.

packages/use-dataloader/src/DataLoaderProvider.tsx
adriengibrat: Ok, it's set twice but with the same constant... not a big deal.

adriengibrat (approved):

vincentaudebert:
Just checking but is the doc/readme up to date with this new param? Could be cool to have some info about it.

vincentaudebert (changes requested):
Just checking but is the doc/readme up to date with this new param? Could be cool to have some info about it.

PR #479, opened 2021-11-08
feat!: support react-router-dom v6

chambo-e:
BREAKING CHANGE: This version is only compatible with react-router-dom v6.


packages/use-query-params/src/__tests__/index.tsx
```diff
@@ -1,30 +1,42 @@
...
 import { History, createMemoryHistory } from 'history'
-import React, { ReactNode } from 'react'
+import React, { ReactNode, useLayoutEffect, useState } from 'react'
 import { MemoryRouter, Router } from 'react-router-dom'
 import useQueryParams from '..'
 
 const wrapper =
-  ({ pathname = 'one', search, history }: { pathname?: string, search: string, history?: History }) =>
+  ({ pathname = '/one', search, history }: { pathname?: string, search: string, history?: History }) =>
   // eslint-disable-next-line react/prop-types
```
QuiiBz, line 9: I tried and it seems like this comment isn't needed anymore

packages/use-query-params/src/__tests__/index.tsx
```diff
@@ -1,30 +1,41 @@
...
+  ({ pathname = '/one', search, history }: { pathname?: string, search: string, history?: History }) =>
+  ({ children }: { children: ReactNode }) => {
+
+    if (history) {
+      const [state, setState] = useState({
+        action: history.action,
+        location: history.location
+      });
+
+      useLayoutEffect(() => history.listen(setState), []);
```
QuiiBz, line 17: Shouldn't hooks be nested in `if` statements?

chambo-e: Only if it affects render, this state is not used outside of the if scope, there should be no issues :)

QuiiBz (approved):

philibea (approved):

PR #486, opened 2021-11-10
chore(node): upgrade node 17 ci

philibea:


chambo-e (approved):
you'll also need to update required workflows in the project settings :)

PR #524, opened 2021-11-30
feat: add paginated dataloader

DorianMaliszewski:
Add a new hook that allow users to handle pagination more easily with our dataloader.


packages/use-dataloader/src/__tests__/usePaginatedDataLoader.test.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+type UseDataLoaderHookProps = {
+  config: UsePaginatedDataLoaderConfig<unknown>
+  key: string
+  method: (params: UsePaginatedDataLoaderMethodParams) => Promise<unknown>
+}
+
+const PROMISE_TIMEOUT = 5
+
+const initialProps = {
```
QuiiBz, line 22: Shouldn't this be typed with `UseDataLoaderHookProps`?

DorianMaliszewski: We don't care because the renderHooks function is typed so it will throw errors if we dont put the right type inside

matthprost (approved):

QuiiBz (approved):

packages/use-dataloader/src/__tests__/usePaginatedDataLoader.test.tsx
```diff
@@ -0,0 +1,213 @@
+/* eslint-disable no-console */
```
philibea, line 1: no console

philibea (approved):

vincentaudebert (approved):

PR #547, opened 2021-12-16
feat(use-i18n): allow default date-fns locale instead of lazy loading

chambo-e:
This is mainly intended when use-i18n is used server side, be it on unit tests or ssr.

On first render the provider will load date-fns locale and then rerender, trigger multiple errors on jest because we don't `act` on render.

When specified a default date-fns locale and no load function provider will not rerender


chambo-e:
Needs https://github.com/scaleway/scaleway-lib/pull/546/files

adriengibrat (approved):

philibea (approved):

PR #564, opened 2021-12-27
feat(eslint-config): enable prefer-object-has-own

chambo-e:


packages/eslint-config-react/shared.js
```diff
@@ -57,6 +59,7 @@ module.exports = {
     // These are rules soon to be enabled by airbnb react config
     // We're getting a head start
     'react/no-adjacent-inline-elements': 'error', // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-adjacent-inline-elements.md
+    'prefer-object-has-own': 'off', // https://eslint.org/docs/rules/prefer-object-has-own
```
vincentaudebert, line 62: off ? Is it to enable or disable the rule?

vincentaudebert (changes requested):

packages/eslint-config-react/shared.js
chambo-e: Uhu, thanks that's fixed 😄

philibea (approved):

vincentaudebert (approved):

PR #598, opened 2022-01-18
feat(segment): add on event error to handle all your events from prov…

philibea:
…ider


packages/use-segment/src/__tests__/index.tsx
```diff
@@ -3,9 +3,22 @@ import { renderHook } from '@testing-library/react-hooks'
...
 import SegmentProvider, { useSegment } from '..'
-import type { Analytics, SegmentProviderProps } from '..'
+import type { Analytics, OnEventError, SegmentProviderProps } from '..'
+
+// type OnEventError = SegmentProviderProps<unknown>['onEventError']
+
+const eventError = new Error('Error Event')
 
 const defaultEvents = {
+  errorEvent: (_?: Analytics, onEventError?: OnEventError) => async () => {
```
chambo-e, line 13: That's not an event, rather an handler :)

DorianMaliszewski (approved):

matthprost (approved):

chambo-e (approved):

PR #612, opened 2022-01-25
fix: lerna publish

DorianMaliszewski:


chambo-e (commented):
Why is this needed ? 🤔

DorianMaliszewski:
> Why is this needed ? 🤔

Needed because of : https://github.com/lerna/lerna/issues/2788#issuecomment-774265338

IK that `npm whoami` is working for automation token but it doesn't work anymore

PR #695, opened 2022-04-01
feat(react): new automatic runtime

philibea:
It's will possibly improve the size on all bundle using react, but we will be more coherent with all ours projects


QuiiBz:
Should we disable the `react/react-in-jsx-scope` rule? ESLint don't like it: 

https://github.com/scaleway/scaleway-lib/runs/5787900573?check_suite_focus=true#step:7:12

philibea:
> Should we disable the `react/react-in-jsx-scope` rule? ESLint don't like it:
> 
> https://github.com/scaleway/scaleway-lib/runs/5787900573?check_suite_focus=true#step:7:12

yes exactly, it's already done

QuiiBz (approved):

packages/use-segment/src/__tests__/index.tsx
```diff
@@ -1,6 +1,6 @@
 import { AnalyticsBrowser, Context } from '@segment/analytics-next'
 import { renderHook } from '@testing-library/react-hooks'
-import React, { ReactNode } from 'react'
+import { ReactNode } from 'react';
```
chambo-e, line 3: intended ?

philibea: fixed with the format https://github.com/scaleway/scaleway-lib/pull/696

chambo-e (approved):

PR #696, opened 2022-04-01
chore(prettier): add pnpm prettignore and format

philibea:
run `pnpm format` 

it's was not run from a long time 🚶🏻


.prettierignore
```diff
@@ -4,3 +4,4 @@ dist/
 
 # Specific to prettier
 package.json
+CHANGELOG.md
```
QuiiBz, line 7: Any reason for not formatting this?

packages/use-query-params/README.md
```diff
@@ -9,10 +9,11 @@ $ pnpm add @scaleway/use-query-params react-router-dom
 ```
 
 ## Usage
+
 ```js
 import React from 'react'
```
QuiiBz, line 14: Is this react import mandatory?

.prettierignore
philibea: it's automatically created by release. it's will make some noise when we want to format

packages/use-query-params/README.md
philibea: We don't know if our user use new jsx transform ?

packages/countries/subdivisions/AD.json
```diff
@@ -48,4 +48,4 @@
     "parent_code": null,
     "type": "Parish"
   }
-]
\ No newline at end of file
+]
```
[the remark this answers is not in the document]
philibea: auto generated by us, we can still apply format ?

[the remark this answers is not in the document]
chambo-e: We need to add json and d.ts files to lint-staged then :)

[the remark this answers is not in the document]
philibea: Rigght, it's done

chambo-e (approved):

QuiiBz (approved):

PR #716, opened 2022-04-11
feat: add useGTM hook

chambo-e:
A new package with a GTM Provider/hook

This allows us to have the same DX as use-segment and consume gtm in a react way

### DIFF
--- a/packages/use-gtm/package.json
+++ b/packages/use-gtm/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@scaleway/use-gtm",
-  "version": "0.1.0",
+  "version": "1.0.0",
   "description": "A small hook to handle gtm in a react app",
   "keywords": [
     "react",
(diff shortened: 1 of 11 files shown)


packages/use-gtm/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/scaleway/scaleway-lib",
+    "directory": "packages/use-gtm"
+  },
+  "license": "MIT",
+  "peerDependencies": {
+    "react": "17.x"
```
QuiiBz, line 30: Maybe?
```suggestion
    "react": "17.x || 18.x"
```

vincentaudebert (approved):
💯

packages/use-gtm/package.json
chambo-e: I think we'll update all packages once we officially support react 18 :)

QuiiBz: Makes sense

QuiiBz (approved):

PR #732, opened 2022-04-21
ci(node): upgrade to node 18

philibea:


QuiiBz:
Looks like node 18 handles `Intl.DateTimeFormat` differently compared to node 16/17

philibea:
> Looks like node 18 handles `Intl.DateTimeFormat` differently compared to node 16/17

yes it's not the first time we have this issue. We can ignore this actually

QuiiBz (approved):

chambo-e:
> > Looks like node 18 handles `Intl.DateTimeFormat` differently compared to node 16/17
> 
> @QuiiBz yes it's not the first time we have this issue. We can ignore this actually

@philibea behaviour is different in node 14,16,17 and 18, we need to find a way to handle this properly

.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        node: [ '14', '16', '17' ]
+        node: ['16', '17', '18' ]
```
chambo-e, line 38: 17 can be removed

Lawndlwd (approved):

chambo-e (approved):

PR #761, opened 2022-05-04
chore(deps): add missing strict peer deps

QuiiBz:
Add missing peer dependencies since pnpm v7 now set `strict-peer-dependencies` to `true`

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@types/jest": "27.4.1",
     "@types/node": "17.0.31",
     "@types/prop-types": "15.7.3",
-    "@types/react": "17.0.39",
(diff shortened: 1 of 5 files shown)


philibea:
@QuiiBz  can you check type https://github.com/scaleway/scaleway-lib/runs/6285324088?check_suite_focus=true

QuiiBz:
> @QuiiBz can you check type https://github.com/scaleway/scaleway-lib/runs/6285324088?check_suite_focus=true

Just pushed, it should be fixed (multiple @types/react versions)

chambo-e (approved):

philibea (approved):

PR #766, opened 2022-05-09
feat: add `use-media` fork to support react 18

QuiiBz:
This library has been forked from [use-media](https://github.com/streamich/use-media), many thanks to the original author, [Vadim Dalecky](https://github.com/streamich).

### DIFF
--- a/packages/use-media/src/__tests__/useMedia.tsx
+++ b/packages/use-media/src/__tests__/useMedia.tsx
@@ -61,7 +61,8 @@ describe('useMedia hook', () => {
     )
 
     unmount()
-    callback?.()
+    // @ts-expect-error variable is assigned inside mockAddEventListener
+    callback()
 
     expect(result.current).toBe(false)
   })


packages/use-media/src/useMedia.ts
```diff
@@ -0,0 +1,51 @@
+import { useEffect, useLayoutEffect, useState } from 'react'
+import { Effect, MediaQueryObject } from './types'
+import { noop, queryObjectToString } from './utilities'
+
+export const mockMediaQueryList: MediaQueryList = {
+  addEventListener: noop,
```
chambo-e, line 6: Could we pass undefined instead of noop ?

packages/use-media/src/utilities/queryObjectToString.ts
```diff
@@ -0,0 +1,30 @@
+import { MediaQueryObject } from '../types'
+import camelToHyphen from './camelToHyphen'
+
+const QUERY_COMBINATOR = ' and '
+
+export default function queryObjectToString(query: string | MediaQueryObject) {
```
chambo-e, line 6: Could we force media query string only ? Less maintenance and better aligned with standards maybe ?

packages/use-media/src/useMedia.ts
QuiiBz: The `MediaQueryList` type doesn't allow `undefined`, so we'll have to stick with a function

packages/use-media/src/utilities/queryObjectToString.ts
QuiiBz: We mostly use `useMedia` with an object property, so we'll need to update it. `MediaQueryObject` doesn't provide useful completions, so I agree that it may be a good idea to only use strings

packages/use-media/src/useMedia.ts
```diff
@@ -0,0 +1,50 @@
...
+      let mounted = true
+      const mediaQueryList: MediaQueryList =
+        typeof window === 'undefined' ||
+        typeof window.matchMedia === 'undefined'
+          ? mockMediaQueryList
+          : window.matchMedia(query)
+
+      const onChange = () => {
+        if (!mounted) {
+          return
```
chambo-e, line 31: Could we find a way to test those use cases ?

chambo-e (changes requested):

DorianMaliszewski (changes requested):
Need 100% coverage :)

chambo-e (approved):

DorianMaliszewski (approved):

PR #786, opened 2022-05-19
feat: dataloader v2

DorianMaliszewski:


packages/use-dataloader/src/dataloader.ts
```diff
@@ -1,248 +1,152 @@
...
-  public maxDataLifetime?: number
-
-  public isDataOutdated = false
+  public static started = 0
 
-  public method: () => PromiseType<T>
+  public static cachedData = {} as Record<string, unknown | undefined>
 
-  private cancelMethod?: () => void
+  public static queue = {} as Record<string, PromiseType>
```
QuiiBz, line 18: Any reason why casting instead of normally typing these variables?

packages/use-dataloader/src/dataloader.ts
```diff
@@ -1,248 +1,152 @@
...
-        }).then(() => this.load(force))
-      }
+  private tryLaunch = async (): Promise<ResultType> => {
+    if (DataLoader.started < DataLoader.maxConcurrent) {
+      DataLoader.started += 1
+      DataLoader.queue[this.key] = this.launch()
+    } else {
+      DataLoader.queue[this.key] = new Promise(resolve => {
+        setTimeout(resolve)
+      }).then(() => this.tryLaunch())
```
QuiiBz, line 72: Not sure if it's easier or not to read
```suggestion
      }).then(this.tryLaunch)
```

packages/use-dataloader/src/dataloader.ts
DorianMaliszewski: No lol

QuiiBz (approved):

packages/use-dataloader/src/useDataLoader.ts
```diff
@@ -1,171 +1,126 @@
...
 import { PromiseType, UseDataLoaderConfig, UseDataLoaderResult } from './types'
 
-/**
- * @param {string} key key to save the data fetched in a local cache
- * @param {() => PromiseType} method a method that return a promise
- * @param {useDataLoaderConfig} config hook configuration
- * @returns {useDataLoaderResult} hook result containing data, request state, and method to reload the data
- */
-const useDataLoader = <T>(
+function useDataLoaderV2<ResultType, ErrorType = Error>(
```
matthprost, line 7: Why renaming in v2 ? Anyways we have to migrate all old ones as there is no coexisting v1 and v2 ?

packages/use-dataloader/src/useDataLoader.ts
```diff
@@ -1,171 +1,126 @@
...
+    return () => {
+      if (timeout) clearTimeout(timeout)
     }
-  })
+  }, [pollingInterval, needPolling, request, request.status, request.data])
 
   return {
-    data: request.getData() ?? (initialData as T),
-    error: request?.error,
+    data: !request.isFirstLoading ? request.data : initialData,
```
matthprost, line 114: I have a tiny improvement here, lets say your initialData is a complex object wouldn't it be nice to spread its default values into data ?

Something like this when request.data and initialData are both objects:
```ts
data: !request.isFirstLoading ? { ...initialData, ...request.data } : initialData,
```

matthprost (changes requested):

matthprost (approved):

PR #865, opened 2022-07-07
feat(use-i18n): add type safety

QuiiBz:
Add type-safe i18n.


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         cache: 'pnpm'
     - run: pnpm install
     - run: pnpm run build
-    - run: pnpm tsc --noEmit
+    - run: pnpm tsc --noEmit --project ./tsconfig.typecheck.json
```
chambo-e, line 19: We should do the opposite no ? Have a custom config for jest only :)

packages/use-i18n/src/types.ts
```diff
@@ -0,0 +1,175 @@
+/**
+ * A locale object which is a record of string:string, with an optional `prefix` with is
```
baptisteArno, line 2:
```suggestion
 * A locale object which is a record of string:string, with an optional `prefix` which is
```

matthprost (approved):

DorianMaliszewski (approved):

PR #880, opened 2022-07-13
fix: use-dataloader types improvement

DorianMaliszewski:


packages/use-dataloader/src/useDataLoader.ts
```diff
@@ -1,10 +1,9 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { useDataLoaderContext } from './DataLoaderProvider'
 import { StatusEnum } from './constants'
-import DataLoader from './dataloader'
 import { PromiseType, UseDataLoaderConfig, UseDataLoaderResult } from './types'
 
-function useDataLoader<ResultType, ErrorType = Error>(
+function useDataLoader<ResultType = unknown, ErrorType = Error>(
```
chambo-e, line 6: Why the default ?

DorianMaliszewski: If we dont care about the result type we can use a default one

chambo-e (approved):

QuiiBz (approved):
LGTM

philibea (approved):

PR #901, opened 2022-08-01
feat: add new regex `alphaDashes` to the list

Lawndlwd:
add new regex `alphaDashes` to the list of regex

Expected to use this regex for name that accept characters and dashes without space on numbers

### DIFF
--- a/packages/use-i18n/src/__tests__/__snapshots__/formatDate.ts.snap
+++ b/packages/use-i18n/src/__tests__/__snapshots__/formatDate.ts.snap
@@ -24,11 +24,11 @@ exports[`formatDate should work with custom format and locale es 2`] = `"jueves,
 
 exports[`formatDate should work with custom format and locale es 3`] = `"jueves, 13 F 20 d. C., 15:28:00 (tiempo universal coordinado)"`;
 
-exports[`formatDate should work with custom format and locale fr 1`] = `"jeudi 13 F 20 ap. J.-C., 16:28:00 temps universel coordonné"`;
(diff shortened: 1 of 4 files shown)


packages/use-i18n/src/__tests__/__snapshots__/formatDate.ts.snap
```diff
@@ -24,45 +24,45 @@ exports[`formatDate should work with custom format and locale es 2`] = `"jueves,
 
 exports[`formatDate should work with custom format and locale es 3`] = `"jueves, 13 F 20 d. C., 15:28:00 (tiempo universal coordinado)"`;
 
-exports[`formatDate should work with custom format and locale fr 1`] = `"jeudi 13 F 20 ap. J.-C., 16:28:00 temps universel coordonné"`;
+exports[`formatDate should work with custom format and locale fr 1`] = `"jeudi 13 F 20 ap. J.-C. 16:28:00 temps universel coordonné"`;
```
chambo-e, line 27: Why ? 🤔

philibea: node 18...

philibea: But this should be updated in this PR ? @Lawndlwd

Lawndlwd: i did` pnpm test` it fails on almost all packages, so it seems it needed snapshot update

Lawndlwd: what i need to do more than `rm -rf node module && pnpm install` to update my pnpm-lock ?

philibea (approved):

matthprost (approved):

PR #920, opened 2022-08-16
feat(use-dataloader): handle primitive array key

philibea:


packages/use-dataloader/src/usePaginatedDataLoader.ts
```diff
@@ -32,9 +32,7 @@ const usePaginatedDataLoader = <ResultType = unknown, ErrorType = Error>(
     perPage = 1,
   }: UsePaginatedDataLoaderConfig<ResultType, ErrorType> = {},
 ): UsePaginatedDataLoaderResult<ResultType, ErrorType> => {
-  if (typeof baseFetchKey !== 'string') {
-    throw new Error(KEY_IS_NOT_STRING_ERROR)
-  }
+  const flatKey = (page: number) => [key, ['page', page]].flat()
```
chambo-e, line 35: need memo here too ?

philibea: can be usefull right

packages/use-dataloader/src/types.ts
```diff
@@ -1,3 +1,6 @@
+type PrimitiveType = string | number | boolean | null | undefined | Date
```
QuiiBz, line 1: I'm not sure why `PrimitiveType` can be `null | undefined`?

packages/use-dataloader/src/usePaginatedDataLoader.ts
```diff
@@ -32,9 +32,10 @@ const usePaginatedDataLoader = <ResultType = unknown, ErrorType = Error>(
     perPage = 1,
   }: UsePaginatedDataLoaderConfig<ResultType, ErrorType> = {},
 ): UsePaginatedDataLoaderResult<ResultType, ErrorType> => {
-  if (typeof baseFetchKey !== 'string') {
-    throw new Error(KEY_IS_NOT_STRING_ERROR)
-  }
+  const flatKey = useMemo(
```
QuiiBz, line 35: Should be a `useCallback` here I think?

QuiiBz (approved):

packages/use-dataloader/src/usePaginatedDataLoader.ts
```diff
@@ -32,9 +32,10 @@ const usePaginatedDataLoader = <ResultType = unknown, ErrorType = Error>(
     perPage = 1,
   }: UsePaginatedDataLoaderConfig<ResultType, ErrorType> = {},
 ): UsePaginatedDataLoaderResult<ResultType, ErrorType> => {
-  if (typeof baseFetchKey !== 'string') {
-    throw new Error(KEY_IS_NOT_STRING_ERROR)
-  }
+  const flatKey = useCallback(
+    (page: number) => [key, ['page', page]].flat(),
+    [key],
+  )
```
chambo-e, line 38: Nope nope, can we memoize the result of `[key, ['page', page]].flat()` ? :)

packages/use-dataloader/src/usePaginatedDataLoader.ts
```diff
@@ -32,13 +32,11 @@ const usePaginatedDataLoader = <ResultType = unknown, ErrorType = Error>(
...
   }: UsePaginatedDataLoaderConfig<ResultType, ErrorType> = {},
 ): UsePaginatedDataLoaderResult<ResultType, ErrorType> => {
-  if (typeof baseFetchKey !== 'string') {
-    throw new Error(KEY_IS_NOT_STRING_ERROR)
-  }
-
   const [data, setData] = useState<Record<number, ResultType | undefined>>({})
   const [page, setPage] = useState<number>(initialPage ?? 1)
 
+  const keyPage = useMemo(() => [key, ['page', page]].flat(), [key, page])
```
chambo-e, line 38:
```suggestion
  const keyPage = useMemo(() => [...key, 'page', page], [key, page])
```

chambo-e (approved):

PR #926, opened 2022-08-18
feat: add regex notHaveSpecialCharacter

Lawndlwd:
Add new regex `notHaveSpecialCharacter`


packages/regex/src/index.ts
```diff
@@ -33,3 +33,4 @@ export const sixDigitsCode = /^[0-9]{6}$/
 export const url =
   /^http(s)?:\/\/?[\w.-]+(?:\.[\w.-]+)+[\w\-._~:/?#[\]@!$&'()*+,;=.]+$/
 export const hexadecimal = /^[0-9a-fA-F]+$/
+export const notHaveSpecialCharacter = /^[^\W][a-zA-Z0-9-_.]+$/
```
chambo-e, line 36: Can we have a regex `specialCharacters` instead ? A negative regex is kinda weird 😄

QuiiBz: Or maybe rename it to `alphaDashDot`?

Lawndlwd: i like `alphanumDashDot`, done

PR #956, opened 2022-08-30
feat(use-i18): typesafe useTranslation

QuiiBz:
Add typesafety on `useTranslation` hook in `@scaleway/use-i18n` and remove prefix.

BREAKING CHANGE: Remove prefix


packages/use-i18n/src/__tests__/usei18n.tsx
```diff
@@ -111,113 +111,6 @@ describe('i18n hook', () => {
     })
   })
 
-  it('should use specific load on useTranslation', async () => {
```
philibea, line 114: why did you removed all test on useTranslation ?

QuiiBz: Because I removed prefix, but it was a mistake. Just added them back without prefix.

matthprost (approved):

baptisteArno (approved):

philibea (approved):

PR #982, opened 2022-09-13
feat(regex): add cidr regex

QuiiBz:
Add cidr regex, from [cidr-regex](https://github.com/silverwind/cidr-regex/blob/master/index.js)


baptisteArno (approved):

matthprost (approved):

chambo-e:
@QuiiBz missing tests

PR #1055, opened 2022-10-25
feat(use-i18n): required generic

QuiiBz:
When using `useI18n()` / `useTranslation()` without providing a generic for the locale, `t` and `namespaceTranslation` will throw an error (`You must pass a generic argument to ****`) to enforce the usage of a Locale.

### DIFF
--- a/packages/use-i18n/src/usei18n.tsx
+++ b/packages/use-i18n/src/usei18n.tsx
@@ -27,7 +27,7 @@ const LOCALE_ITEM_STORAGE = 'locale'
 type TranslationsByLocales = Record<string, BaseLocale>
 type RequiredGenericContext<Locale extends BaseLocale> =
   keyof Locale extends never
-    ? Context<Locale> & {
+    ? Omit<Context<Locale>, 't' | 'namespaceTranslation'> & {
         t: (str: 'You must pass a generic argument to useI18n()') => void
         namespaceTranslation: (
           str: 'You must pass a generic argument to useI18n()',
--- a/packages/use-i18n/src/usei18n.tsx
+++ b/packages/use-i18n/src/usei18n.tsx
@@ -25,6 +25,15 @@ import type { ReactParamsObject, ScopedTranslateFn, TranslateFn } from './types'
 const LOCALE_ITEM_STORAGE = 'locale'
 
 type TranslationsByLocales = Record<string, BaseLocale>
+type RequiredGenericContext<Locale extends BaseLocale> =
+  keyof Locale extends never
+    ? Context<Locale> & {
+        t: (str: 'You must pass a generic argument to useI18n()') => void
+        namespaceTranslation: (
+          str: 'You must pass a generic argument to useI18n()',
+        ) => void
+      }
+    : Context<Locale>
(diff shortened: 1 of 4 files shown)


packages/use-i18n/src/usei18n.tsx
```diff
@@ -99,19 +108,23 @@ interface Context<Locale extends BaseLocale> {
...
   if (context === undefined) {
     throw new Error('useI18n must be used within a I18nProvider')
   }
 
-  return context as unknown as Context<Locale>
+  return context as unknown as RequiredGenericContext<Locale>
 }
 
-export function useTranslation<Locale extends BaseLocale>(
+// eslint-disable-next-line @typescript-eslint/ban-types
```
chambo-e, line 123: which type is banned ?

chambo-e (changes requested):

packages/use-i18n/src/usei18n.tsx
QuiiBz: `{}`

packages/use-i18n/src/usei18n.tsx
```diff
@@ -99,19 +108,23 @@ interface Context<Locale extends BaseLocale> {
 // useI18n / useTranslation requires to explicitely give a Locale to use.
 const I18nContext = createContext<Context<any> | undefined>(undefined)
 
-export function useI18n<Locale extends BaseLocale>(): Context<Locale> {
+export function useI18n<
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  Locale extends BaseLocale = {},
```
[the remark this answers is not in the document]
QuiiBz: Because there is no other way to detect if the user provided a generic or not. We have to set is to an empty object, and if this objects doesn't have any keys (`keyof Locale extends never`) we return the errors.

chambo-e (approved):

packages/use-i18n/src/usei18n.tsx
chambo-e: Why not use object ? Or Record ?

QuiiBz: Because it has to be an `{}` to check if it's empty (= `keyof Locale extends never`) or not in order to show the error message. This [this playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UDWEQHsBmUDeBfAUFFAPTFQB6A-PvqJFAEILKqZQDyARgFYQDGwQiTJUhtcNADCzFOiwAlfmgBOAEwA8AZ2DKAlgDsA5gBoo+iADcIygHxDSFSkA)

matthprost (approved):

packages/use-i18n/src/usei18n.tsx
chambo-e: Screenshot 2022-10-26 at 11 49 56

QuiiBz: `object` by itself doesn't satisfy the constraint of `BaseLocale`, because it can be any object but we require an object with keys being strings.

PR #1079, opened 2022-11-17
fix: cron regex

vincentaudebert:
Correct cron regex


chambo-e (changes requested):
Could you also add some tests matching the new regex ? :)

QuiiBz (approved):

vincentaudebert:
done @chambo-e

PR #1089, opened 2022-11-25
feat: introduce @scaleway/use-storage

chambo-e:


packages/use-storage/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+  const setValue = useCallback(
+    (val: T | undefined) => {
+      if (val) {
+        if (canUseDOM) {
+          flushSync(() => storage.setItem(key, JSON.stringify(val)))
+        }
+        setLocaleValue(JSON.stringify(val))
+      } else {
+        if (canUseDOM) {
+          flushSync(() => storage.removeItem(key))
```
QuiiBz, line 73: `storage.setItem()` / `storage.getItem()` doesn't trigger any React state update, so `flushSync` is useless (unless I'm missing something?)

packages/use-storage/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Usage
+
+With localStorage
+
+```js
+import React from 'react'
+import { useLocalStorage } from '@scaleway/use-storage'
+
+const Component = () => {
+  const [value, setValue] = useLocalStorage('key')
```
QuiiBz, line 22: Could we add an example with `initialValue`?

packages/use-storage/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+  typeof window !== 'undefined' &&
+  typeof window?.document !== 'undefined' &&
+  typeof window?.localStorage !== 'undefined' &&
+  typeof window?.sessionStorage !== 'undefined'
+)
+
+const subscribeStorage = (callback: () => void) => {
+  if (canUseDOM) {
+    window.addEventListener('storage', callback)
+    window.addEventListener('event-storage', callback)
```
QuiiBz, line 26: Where does this `event-storage` comes from?

QuiiBz (changes requested):

packages/use-storage/src/index.ts
chambo-e: True, especially since we move to a subscription based state

packages/use-storage/src/index.ts
chambo-e: From the setValue below which trigger this custom event

QuiiBz: Why not use the existing `storage` event instead of creating another and listening twice for the same thing?

chambo-e: Because `storage` is not broadcasted to current tab :)

https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

QuiiBz: Alright, could you add a comment to explain this?

chambo-e: Done :)

QuiiBz (approved):
LGTM

fabienhebert (approved):

PR #1093, opened 2022-11-28
feat(regex): add s3 bucket name regex

QuiiBz:
Add s3 bucket name regex


packages/regex/src/index.ts
```diff
@@ -36,6 +36,7 @@ export const sixDigitsCode = /^[0-9]{6}$/
 export const url =
   /^http(s)?:\/\/?[\w.-]+(?:\.[\w.-]+)+[\w\-._~:/?#[\]@!$&'()*+,;=.]+$/
 export const hexadecimal = /^[0-9a-fA-F]+$/
+export const s3BucketName = /^[a-z0-9._-]{3,}$/
```
chambo-e, line 39: Still needed with the new hive backend ? 🤔

QuiiBz: It's part of the S3 spec, so yes

chambo-e (approved):

ModuloM (approved):

PR #1128, opened 2022-12-21
fix(use-i18n): correct formatUnit type

chambo-e:
Type was not accepting `megabit-per-second` while the code supports it

### DIFF
--- a/packages/use-i18n/src/__tests__/formatUnit.ts
+++ b/packages/use-i18n/src/__tests__/formatUnit.ts
@@ -5,45 +5,45 @@ const locales = ['en', 'fr', 'ro']
 
 type TestType = [string, FormatUnitOptions, string, number]
 
+const SUPPORTED_UNITS = Object.keys(
+  supportedUnits,
+) as FormatUnitOptions['unit'][]
+
 const tests: TestType[] = [
-  ...(Object.keys(supportedUnits).map(unit => [
+  ...SUPPORTED_UNITS.map<TestType>(unit => [
     'should work with',
     { unit },
     'fr',
     12.56,
-  ]) as TestType[]),
-  ...(Object.keys(supportedUnits)
-    .map(unit =>
-      locales.map(locale => [
-        `should work with locale ${locale} and`,
-        { unit },
-        locale,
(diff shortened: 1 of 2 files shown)


ModuloM (approved):

packages/use-i18n/src/__tests__/formatUnit.ts
```diff
@@ -3,14 +3,16 @@ import formatUnit, { supportedUnits } from '../formatUnit'
...
+
+const tests: TestType[] = [
+  ...(Object.keys(supportedUnits).map(unit => [
     'should work with',
     { unit },
     'fr',
     12.56,
-  ]),
-  ...Object.keys(supportedUnits)
+  ]) as TestType[]),
```
QuiiBz, line 14: It's better to use a generic inside `map` instead of casting everywhere:

```ts
const KEYS = Object.keys(supportedUnits) as SupportedUnits[]

const tests: TestType[] = [
  ...(KEYS.map<TestType>(unit => [
    'should work with',
    { unit },
    'fr',
    12.56,
  ])),
  ...(KEYS
    .map(unit =>
      locales.map<TestType>(locale => [
        `should work with locale ${locale} and`,
        { unit },
        locale,
        12.56,
      ]),
    )
    .flat()),
  ...(KEYS
    .map(unit =>
      locales.map<TestType>(locale => [
        `should work with long format, locale ${locale} and`,
        { short: false, unit },
        locale,
        12.56,
      ]),
    )
    .flat()),
  ...(KEYS.map<TestType>(unit => [
    'should work with maximumFractionDigits',
    { maximumFractionDigits: 3, unit },
    'fr',
    12.56,
  ])),
  ...(KEYS.map<TestType>(unit => [
    'should work with minimumFractionDigits',
    { minimumFractionDigits: 3, unit },
    'fr',
    12,
  ])),
]
```

chambo-e: Roh, perfect, couldn't manage to make it work but yours is ok, thx :)

QuiiBz (approved):

PR #1140, opened 2023-01-04
fix(prop-types): remove usage of proptypes

philibea:


chambo-e (approved):

packages/use-dataloader/src/DataLoaderProvider.tsx
```diff
@@ -59,8 +58,8 @@ export const DataLoaderContext = createContext<IDataLoaderContext>(undefined)
 const DataLoaderProvider = ({
   children,
   cacheKeyPrefix,
-  onError,
-  maxConcurrentRequests,
+  onError = undefined,
```
QuiiBz, line 61: The type should be `onError?: OnErrorFn` and then we can remove this `= undefined`

philibea: yes i already made the change ;)

QuiiBz (approved):

PR #1145, opened 2023-01-06
feat!: publish packages as ESM only

chambo-e:
BREAKING CHANGE: packages are ESM only

### DIFF
--- a/packages/jest-helpers/package.json
+++ b/packages/jest-helpers/package.json
@@ -3,9 +3,13 @@
   "version": "2.0.25",
   "description": "A package for utilities jest functions",
   "type": "module",
+  "engines": {
+    "node": ">=14.x"
+  },
+  "sideEffects": false,
   "main": "dist/index.js",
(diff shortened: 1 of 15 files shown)


packages/jest-helpers/package.json
```diff
@@ -3,9 +3,13 @@
...
   "description": "A package for utilities jest functions",
   "type": "module",
+  "engines": {
+    "node": ">=14.x"
+  },
+  "sideEffects": false,
   "main": "dist/index.js",
   "module": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "jsnext:main": "dist/index.js",
```
QuiiBz, line 12: > jsnext:main has been superseded by pkg.module, which indicates the location of a file with import/export declarations.

https://github.com/jsforum/jsforum/issues/5

Also we should keep the `types` field, not sure why you removed it?

matthprost (approved):

QuiiBz (approved):

PR #1258, opened 2023-03-24
fix(use-dataloader): fix cache array key with false value

ModuloM:


philibea (approved):

packages/use-dataloader/src/helpers.ts
```diff
@@ -8,7 +8,7 @@ import type { KeyType } from './types'
 export const marshalQueryKey = (queryKey: KeyType) =>
   Array.isArray(queryKey)
     ? queryKey
-        .filter(Boolean)
+        .filter(value => value !== undefined)
```
chambo-e, line 11:
```suggestion
        .filter(value => value !== undefined && value !== null)
```

chambo-e (changes requested):

packages/use-dataloader/src/helpers.ts
ModuloM: Not sure it is safe to do this, same thing with `false` value, the cache key should be different when value is undefined and then null. But maybe I'm missing a test for this.

ModuloM: Done, with comment

PR #1276, opened 2023-04-06
fix: enableDebugKey display key

DorianMaliszewski:
enableDebugKey should display key instead of missing key


ModuloM (approved):
Manque changelog ;)

chambo-e (approved):

philibea (approved):

PR #1289, opened 2023-04-26
ci(npm): add provenance configuration

matthprost:
Added configuration to integrate provenance certification on NPM.


QuiiBz (approved):

.github/workflows/release.yml
```diff
@@ -18,7 +19,7 @@ jobs:
           git config --global user.email 'github@scaleway.com'
           git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/$GITHUB_REPOSITORY
         env:
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
chambo-e, line 22: Why ? These are not the same, GH_TOKEN is our token 😄

matthprost: @chambo-e why not using injected one by github?

chambo-e: It's a scalway-bot account PAT, so we'd release with it

But that does not seems to have any effect now so 🤷

chambo-e (approved):

PR #1320, opened 2023-05-19
feat(use-i18n): make the provider compatible with server side rendering

ModuloM:
Verify that the provider renders client side before checking locale from either `navigator` or `localSotrage`.

### DIFF
--- a/packages/use-i18n/src/__tests__/usei18n.tsx
+++ b/packages/use-i18n/src/__tests__/usei18n.tsx
@@ -237,7 +237,7 @@ describe('i18n hook', () => {
     })
   })
 
-  it('should set current locale from navigator languages', async () => {
+  it('should set current locale from defaultLocale', async () => {
     const { result } = renderHook(() => useI18n(), {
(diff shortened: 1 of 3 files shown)


packages/use-i18n/src/__tests__/usei18n.tsx
```diff
@@ -250,6 +250,89 @@ describe('i18n hook', () => {
...
+      const { result } = renderHook(() => useI18n(), {
+        wrapper: wrapper({
+          defaultLocale: 'es',
+          supportedLocales: ['en', 'fr', 'es'],
+        }),
+      })
+
+      await waitFor(() => {
+        expect(result.current.currentLocale).toEqual('en')
+      })
```
QuiiBz, line 277: Could we add `expect`s for the above localStorage mocks to make sure the proper method is called/not called?

QuiiBz (approved):

Lawndlwd (approved):

Lawndlwd (approved):

PR #1405, opened 2023-07-26
fix(formatUnit): allow to change base on formatUnit

vincentaudebert:
Allow to specify a base (other than 10) on formatUnit function.

This way 4294967296 can format to 4 GB for example.

### DIFF
--- a/packages/use-i18n/src/formatUnit.ts
+++ b/packages/use-i18n/src/formatUnit.ts
@@ -137,10 +137,12 @@ const format =
       maximumFractionDigits,
       minimumFractionDigits,
       short = true,
+      base = 10,
     }: {
       maximumFractionDigits?: number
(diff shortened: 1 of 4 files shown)


QuiiBz (approved):

packages/use-i18n/src/formatUnit.ts
```diff
@@ -248,6 +250,9 @@ export const supportedUnits: Partial<
 export type FormatUnitOptions = {
   unit: SupportedUnits
   short?: boolean
+  maximumFractionDigits?: number
+  minimumFractionDigits?: number
+  base?: number
```
philibea, line 255: we can add a comment to say that the default base will be 10

chambo-e (approved):

PR #1416, opened 2023-07-31
feat(growthbook): add new package to mutualise GrowthBook init

Slashgear:
## Why

We want to add AB test and feature flag frontend tools to many projects at scaleway.
In order to lever what have been done in Console project, I propose to extract the `init` of AB test in a dedicate package which has the responsability and API of it.

For now, [Growthbook](https://docs.growthbook.io/lib/react) is used but this package allow to change it or to enhance/adapt it to our usage and share it within the company.

## How

- create a dedicate package
- add README to describe the usage of it
- Ignore Ts errors as GrowthBook JS sdk has some error with the Typescrypt definition export.
- Facade it to ABTest API (we don't know if growthbook will be their forever)


philibea:
@Slashgear  missing some file inside the package: 
- .npmignore
- .eslintrc.cjs

packages/growthbook/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@scaleway/growthbook",
```
philibea, line 2: To be align with ours packages, i think it's `@scaleway/use-growthbook`

philibea:
@Slashgear missing some test ?

packages/use-growthbook/src/AbTestProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-expect-error TODO: remove once Growthbook is correctly typed and export
+import { GrowthBook, GrowthBookProvider } from '@growthbook/growthbook-react'
+import type { ReactNode } from 'react'
+import { useCallback, useEffect, useMemo } from 'react'
+import type { GrowthBookType } from './types'
+
+export type ToolConfig = {
```
chambo-e, line 12: Can we pick type from Growthbook here ?

packages/use-growthbook/src/AbTestProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+    enableDevMode,
+    attributes: {
+      anonymousId,
+      userId: undefined,
+      organizationId: undefined,
+      organizationType: undefined,
+    },
+    trackingCallback,
+  })
+export function AbTestProvider({
```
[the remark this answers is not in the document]
Slashgear: I saw both on the monorepo so I didn't know what to choose. Any recommendation ?

packages/use-growthbook/src/AbTestProvider.tsx
Slashgear: You mean official typing from growthbook or do you mean moving this type in the types.ts file ?

packages/use-growthbook/src/AbTestProvider.tsx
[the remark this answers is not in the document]
chambo-e: IIRC function were used because typescript didn't support generic const function declaration, that has been fixed on their side a while ago, I'll try to convert them to const :)

But yeah, we favour const anywhere we can

chambo-e (approved):

QuiiBz (approved):

philibea (approved):

PR #1448, opened 2023-08-10
feat(regex): add subdomain regex

johnrazeur:
Add a new subdomain regex.


packages/regex/src/__tests__/index.ts
```diff
@@ -41,6 +41,7 @@ import {
   sixDigitsCode,
   spaces,
   uppercaseBasicDomain,
+  uppercaseBasicSubdomain,
```
chambo-e, line 44: What is a subdomain ? Only xxx.foo.tld ? Or can be nested like (zzz*n).yyy.xxx.foo.tld ?

johnrazeur: Yes it can be nested.

chambo-e (approved):

philibea (approved):

PR #1494, opened 2023-09-01
feat(cookie): init cookie consent package

Axel-KIRK:
## What
- Extract internal cookie consent provider strategy to shared package


packages/cookie-consent/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## QuickStart
+
+### Prerequisites
+
+```
+$ cd packages/cookie-consent
+```
+
+### Start
```
chambo-e, line 13: Can we add an example of how to use it ?

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
```diff
@@ -0,0 +1,227 @@
...
+  useEffect,
+  useMemo,
+  useState,
+} from 'react'
+import { uniq } from '../helpers/array'
+import { stringToHash } from '../helpers/misc'
+import type { CategoryKind, Config, Consent, Integrations } from './types'
+import { useSegmentIntegrations } from './useSegmentIntegrations'
+
+const COOKIE_PREFIX = '_scw_rgpd'
```
chambo-e, line 16: Should this be customizable by config ?

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
```diff
@@ -0,0 +1,227 @@
...
+const HASH_COOKIE = `${COOKIE_PREFIX}_hash`
+
+// Appx 13 Months
+const CONSENT_MAX_AGE = 13 * 30 * 24 * 60 * 60
+// Appx 6 Months
+const CONSENT_ADVERTISING_MAX_AGE = 6 * 30 * 24 * 60 * 60
```
chambo-e, line 22: Same for these, should be customizable no ?

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+const COOKIES_OPTIONS = {
```
chambo-e, line 24: Also those 😄

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
```diff
@@ -0,0 +1,227 @@
...
+    }),
+    [cookieConsent, segmentIntegrations],
+  )
+
+  const value = useMemo(
+    () => ({
+      integrations,
+      needConsent,
+      isSegmentAllowed,
+      segmentIntegrations: segmentEnabledIntegrations,
```
chambo-e, line 208: Don't think we need this in the context anymore

packages/cookie-consent/src/CookieConsentProvider/types.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export type Integrations = Integration[]
+
+// TODO: avoid duplicating this type in @shire/console
+export type Config = {
+  segment?: {
+    writeKey: string
+    cdnURL: string
+  } | null
+}
```
chambo-e, line 20: Can be passed as props instead of mimicking the console config object :)

Also no need to mention the console project, should be fully unrelated

packages/cookie-consent/src/helpers/array.ts
```diff
@@ -0,0 +1,2 @@
+// TODO: avoid duplicating this function in @shire/console
```
chambo-e, line 1: No need :)

Same for the file below

Slashgear: Yes

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
Slashgear: At first we could only mirror exactly what we have now and then add new features.
We don't expect for people to use this package, we only intend to make it public

But we could do some changes yes.

Personnally I would have put them as default value in order to ease the integration of this external package and limit risks

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
Slashgear: It is still used in console.

packages/cookie-consent/src/CookieConsentProvider/useSegmentIntegrations.ts
```diff
@@ -0,0 +1,86 @@
...
+): Integrations =>
+  [defaultSegmentIoIntegration, ...segmentIntegrations].map(
+    ({ name, category, creationName }) => ({
+      // Segment requires the `creationName` for this destination.
+      // This condition is a test (as of 2023-02-28)
+      // and should either be improved or deleted.
+      name: name === 'Google Ads (Gtag)' ? creationName : name,
+      category: CATEGORY_MATCH[category] ?? 'marketing',
+    }),
+  )
```
Vinzius, line 47: We initially added this condition to verify the creationName was the proper slug to send back to Segment. IMO we should simplify and always send the creationName, which is the recommended way by Segment.

```suggestion
  [defaultSegmentIoIntegration, ...segmentIntegrations].map(
    ({ category, creationName }) => ({
      name: creationName,
      category: CATEGORY_MATCH[category] ?? 'marketing',
    }),
  )
```

Would need to be monitor, but I guess it would be a good timing if we include this package.

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
```diff
@@ -0,0 +1,245 @@
...
+  essentialIntegrations: string[]
+  config: Config
+  cookiePrefix: string
+  consentMaxAge: number
+  consentAdvertisingMaxAge: number
+  cookiesOptions: {
+    sameSite: boolean | 'strict' | 'lax' | 'none' | undefined
+    secure: boolean
+    path: string
+  }
```
QuiiBz, line 73: Could we derive this type from `COOKIES_OPTIONS`, or create a separate type to be used in both places?

QuiiBz (changes requested):

packages/cookie-consent/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@scaleway/cookie-consent",
+  "version": "0.2.14",
```
Vinzius, line 3: Question: why using with this value, and not starting with 0.1 or similar?

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
```diff
@@ -0,0 +1,245 @@
...
+  sameSite: 'strict',
+  secure: true,
+  path: '/',
+} as const
+
+type Context = {
+  integrations: Integrations
+  needConsent: boolean
+  isSegmentAllowed: boolean
+  segmentIntegrations: { All: boolean } & Record<string, boolean>
```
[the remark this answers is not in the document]
Slashgear: Don't really know, it was like that at first, we could change that but that will imply a breaking change in parameters for the one that use this attributes.

packages/cookie-consent/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "Cookie consentment banner",
```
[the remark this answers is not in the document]
Slashgear: For sure !

packages/cookie-consent/src/helpers/misc.ts
```diff
@@ -0,0 +1,2 @@
+export const stringToHash = (str = ''): number =>
+  Array.from(str).reduce((s, c) => Math.imul(31, s) + c.charCodeAt(0) || 0, 0)
```
Vinzius, line 2: Question/nit: y-a-t-il un intérêt d'avoir un input vide par défaut ? On pourrait l'enlever (et du coup typer par string ?)

Vinzius (commented):
Thanks Team, for tackling this task ^^

packages/cookie-consent/package.json
Slashgear: Copy paste issue, I am on it ;)

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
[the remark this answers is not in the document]
QuiiBz: It's a new package, so IMO it's fine to update the API if that means we fix some small quirks at the same time.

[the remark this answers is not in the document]
Slashgear: I am on it then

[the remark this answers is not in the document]
Slashgear: Ah I understood why it is in PascalCase...

SegmentIntegration ofbject contains Record<IntegrationName, status>
Integration names are in Pascal case, it seems it feel more natural

```diff
 Object {
    -   "All": false,
        "Google Universal Analytics": false,
        "Salesforce": true,
        "Salesforce custom destination (Scaleway)": true,
    +   "all": false,
      }
```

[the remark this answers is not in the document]
QuiiBz: Oh ok makes sense, then we can keep it like that. Thanks!

Slashgear (approved):

QuiiBz (approved):

Vinzius (approved):

PR #1514, opened 2023-09-13
chore(random-name): use backward compatible module and types configuration

quantumsheep:
The current way of resolving modules is resulting in issues with eslint ([example](https://github.com/scaleway/scaleway-sdk-js/actions/runs/6169349622/job/16743432874?pr=819)).

Setting the module and types path the same way as `@scaleway/configuration-loader` is working (it's backward compatible).


quantumsheep:
Cancelled because that's the way we should do things

PR #1553, opened 2023-10-05
feat(cookie-consent): pass cookie options when deleting one consent

rachidBensaid:
## Why

We were having an issue on localhost when trying to remove a cookie. It wasn't always working, apparently we need to pass the path (cookie options in our case) to make it work. You can have more info by clicking on the lint below

https://stackoverflow.com/a/28119715


Slashgear (approved):
LGTM, nice 🎉

philibea (approved):

PR #1599, opened 2023-10-23
docs(readme): fixup links and bundlephobias

Slashgear:
## Why

Some links are broken in the global readme of the repository
- bundlephobia pages are not working because package does not export entrypoints
- they are broken links in growthbook package

## How

- fix global readme links
- re-add main entry to make bundlephobia working

@QuiiBz What is your opinion on re-adding main key on package ?

### DIFF
--- a/.changeset/rotten-beans-juggle.md
+++ b/.changeset/rotten-beans-juggle.md
@@ -1,6 +1,0 @@
----
----
-
-- fix links of growthbook package
-- replace usage of bundlephobia by packagephobia
-
--- a/.changeset/rotten-beans-juggle.md
+++ b/.changeset/rotten-beans-juggle.md
@@ -1,0 +1,6 @@
+---
+---
+
+- fix links of growthbook package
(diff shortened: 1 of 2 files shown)


QuiiBz (changes requested):
We shouldn't (and don't) need the `main` field in each `package.json`, since we're using the "new" `exports` field instead which covers all our needs.

It's probably a bug (or at least an unsupported feature) of bundlephobia to not support the `exports` field - example with packagephobia, which supports it fine: https://packagephobia.com/result?p=%40scaleway%2Fjest-helpers

Slashgear:
Ok I will change that

.changeset/rotten-beans-juggle.md
QuiiBz, line 1: We should probably remove the changeset since no packages are impacted.

QuiiBz (approved):

ModuloM (approved):

PR #1605, opened 2023-10-26
fix: absolute path rule

DorianMaliszewski:


ModuloM (approved):

philibea:
@DorianMaliszewski  missing changeset, and also test u on node 18 is not working.

philibea (approved):

PR #1690, opened 2023-11-26
refactor(type-test): use new runner for types tests

philibea:
Migration from "tsd-lite" to "tstyche".

waiting https://github.com/tstyche/tstyche/issues/17

close https://github.com/scaleway/scaleway-lib/pull/1658

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,15 +52,13 @@ jobs:
       - run: pnpm install
       - run: pnpm run build
       - run: pnpm run test:coverage
-<<<<<<< HEAD
       - uses: codecov/codecov-action@v4.1.1
-=======
-      - uses: codecov/codecov-action@v3.1.4
+
   test-types:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        node: ['18','20']
+        node: ['20']
     steps:
       - uses: actions/checkout@v4.1.1
       - uses: pnpm/action-setup@v2.4.0
@@ -74,4 +72,3 @@ jobs:
       - run: pnpm run build
       - run: pnpm run test:types
 
->>>>>>> e23dfe49 (refactor(type-test): use new runner for types tests)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,4 +52,26 @@ jobs:
       - run: pnpm install
       - run: pnpm run build
       - run: pnpm run test:coverage
+<<<<<<< HEAD
       - uses: codecov/codecov-action@v4.1.1
+=======
+      - uses: codecov/codecov-action@v3.1.4
+  test-types:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        node: ['18','20']
+    steps:
+      - uses: actions/checkout@v4.1.1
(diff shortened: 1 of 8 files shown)


philibea:
Waiting https://github.com/tstyche/tstyche/issues/17

mrazauskas:
Thanks for giving TSTyche a try. I was wondering why did you marked this PR as blocked by https://github.com/tstyche/tstyche/issues/17?

In a way current `tstyche` matchers can be used as their `tsd` equivalents. I agreed that `.toBeCallable()` would be perfect here. Just trying to say that it also can be adopted gradually some time later. `.toBeCallable()` will take time to get implement. I will make it work, but this is rather involving.

It was good idea to wait a bit, of course. TSTyche became more performant, because the latest release by default will load `typescript` package from your the repo. This is big improvement, because, as you know, `jest-runner-tsd` requires `@tsd/typescript` to be installed additionally.

I was simply curious to see how TSTyche will work with `pnpm`. Hoping to convince you to migrate soon (; Let me know, if there is anything I can help.

package.json
```diff
@@ -52,7 +52,7 @@
     "rollup-plugin-dts": "6.1.0",
     "rollup-plugin-preserve-shebangs": "0.2.0",
     "rollup-plugin-visualizer": "5.11.0",
-    "tsd-lite": "0.8.0",
+    "tstyche": "1.0.0-beta.6",
```
mrazauskas, line 55: You can also remove `@tsd/typescript` and `jest-runner-tsd`.

tstyche.config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "testFileMatch": ["**/__typetests__/*.test.ts", "**/__typetests__/*.test.tsx"]
```
mrazauskas, line 2: Hm.. The default `"**/__typetests__/*.test.*"` pattern did not work? Sounds you don’t need `tstyche.config.json`.

.github/workflows/ci.yml
```diff
@@ -53,3 +53,21 @@ jobs:
       - run: pnpm run build
       - run: pnpm run test:coverage
       - uses: codecov/codecov-action@v3.1.4
+  test-types:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        node: ['18','20']
```
mrazauskas, line 60: Not sure if Node.js versions makes any difference. Perhaps? In a way TypeScript should work the same way on all supported Node.js versions.

Did you consider `tstyche --target 5.0,current` to test types using TypeScript 5.0 and currently version? This is useful in if you want to check that TypeScript 5.0 (or any other lowest supported version) really works.

package.json
philibea: Perfect, thanks

tstyche.config.json
philibea: Yes your right

.github/workflows/ci.yml
philibea: It's was our default configuration in case there is any issue with the next node.js lts version. Thanks for taking the time to review and sorry to come back so late.

And maybe we will try to add a matrix of TypeScript version later if it worth it.

PR #1722, opened 2023-12-12
fix(useSegment): add ready state

ModuloM:
Add ready state to prevent calling events from Provider when analytics instance is not ready

### DIFF
--- a/packages/use-segment/src/useSegment.tsx
+++ b/packages/use-segment/src/useSegment.tsx
@@ -41,6 +41,7 @@ export function useSegment<T extends Events>(): SegmentContextInterface<T> {
 export type SegmentProviderProps<T> = {
   settings?: AnalyticsBrowserSettings
   initOptions?: InitOptions
+  areOptionsLoaded?: boolean
   onError?: (err: Error) => void
   onEventError?: OnEventError
   events: T
@@ -53,6 +54,7 @@ function SegmentProvider<T extends Events>({
   children,
   settings,
   initOptions,
+  areOptionsLoaded = false,
   onError,
   onEventError,
   events,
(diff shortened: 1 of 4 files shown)


packages/use-segment/src/useSegment.tsx
```diff
@@ -56,6 +57,8 @@ function SegmentProvider<T extends Events>({
   onEventError,
   events,
 }: SegmentProviderProps<T>) {
+  // FIXME set states are triggering a warning when login with magic link
```
philibea, line 60: @ModuloM it's a public repository, i don't think this is usefull

packages/use-segment/src/useSegment.tsx
```diff
@@ -4,7 +4,7 @@ import type {
   AnalyticsBrowserSettings,
   InitOptions,
 } from '@segment/analytics-next'
-import type { ReactNode } from 'react'
```
DorianMaliszewski, line 7: I dont like this syntax where type is inside :/

BABAK0T0 (approved):

JulienSaguez (approved):

packages/use-segment/src/useSegment.tsx
```diff
@@ -52,34 +54,49 @@ function SegmentProvider<T extends Events>({
...
 
   const shouldLoad = useMemo(() => {
+    // If options are loading (for example Segment integrations),
+    // we don't know if we should load or not
+    if (areOptionsLoading) {
+      return undefined
+    }
+
     const hasNoIntegrationsSettings = !initOptions?.integrations
-    const isAllEnabled = !!initOptions?.integrations?.All
```
philibea, line 65: I think it's still use full and should not be removed, as we can approved All integration do you know if one of the integration will be set to True ? if not isAnyIntegrationEnabled will be false also and nothing be load, just check in case it's happen

packages/use-segment/src/useSegment.tsx
```diff
@@ -52,34 +54,49 @@ function SegmentProvider<T extends Events>({
...
       initOptions?.integrations ?? {},
     ).reduce<boolean>((acc, integration) => !!acc || !!integration, false)
 
     return (
       !!settings?.writeKey &&
-      (hasNoIntegrationsSettings || isAllEnabled || isAnyIntegrationEnabled)
+      (hasNoIntegrationsSettings || isAnyIntegrationEnabled)
     )
-  }, [initOptions?.integrations, settings?.writeKey])
+  }, [areOptionsLoading, initOptions, settings])
```
philibea, line 83:
```suggestion
  }, [areOptionsLoading, initOptions?.integrations, settings?.writeKey])
```

philibea: This will re-render everytime the object change not what you trutly use ?

packages/use-segment/src/useSegment.tsx
```diff
@@ -52,34 +54,49 @@ function SegmentProvider<T extends Events>({
   children,
   settings,
   initOptions,
+  areOptionsLoading = false,
```
philibea, line 57: should we check if areOptionLoaded instead of isLoading as with loading state we can have a false, true, false sequency ? 
and with areOptionLoaded this is, false => true  ?

packages/use-segment/src/useSegment.tsx
ModuloM: Good catch, my IDE is somehow too optimistic ;)

packages/use-segment/src/useSegment.tsx
ModuloM: I've tested it in integration with the useCookieConsent and it didn't append (because i've passed in the isAnalyticsLoading state that is true by default). But we need to have a default to avoid breaking change no? (and prevent a major release)

ModuloM: Maybe I didn't get your point thought.

philibea (approved):

PR #1782, opened 2024-01-15
fix(i18n): add onTranslateError and fallback on defaultLocal with con…

philibea:
…text value

### DIFF
--- a/packages/use-i18n/src/usei18n.tsx
+++ b/packages/use-i18n/src/usei18n.tsx
@@ -214,7 +214,17 @@ const I18nContextProvider = ({
   enableDebugKey: boolean
   localeItemStorage: string
   supportedLocales: string[]
-  onTranslateError?: (error: Error) => void
+  onTranslateError?: ({
+    error,
+    currentLocale,
+    value,
+    key,
+  }: {
+    error: Error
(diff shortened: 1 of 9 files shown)


packages/use-i18n/src/usei18n.tsx
```diff
@@ -390,14 +393,31 @@ const I18nContextProvider = ({
...
       if (context) {
-        return formatters
-          .getTranslationFormat(value, currentLocale)
-          .format(context) as string
+        try {
+          return formatters
+            .getTranslationFormat(value, currentLocale)
+            .format(context) as string
+        } catch (err) {
+          onTranslateError?.(err as Error)
```
fabienhebert, line 401: Sounds great, may be we could provide `currentLocale` as string  to quickly find which locale file the is related  to the issue

fabienhebert (approved):

packages/use-i18n/src/usei18n.tsx
philibea: Exact, something in mind https://github.com/scaleway/scaleway-lib/pull/1747

ModuloM (approved):

PR #1902, opened 2024-04-01
fix(integrations): do not set All by default when integrations are al…

philibea:
Remove an Objet All: false when there is more than one integration as we set all categories actully.

Fix test unitaire, test was not properly clear between them.

Avoid unecessary re-redender with a useless state of cookies


Slashgear (approved):
Not a big fan of the `undefined` passed in function sort and join but ok

PR #2041, opened 2024-06-13
fix(cookie-consent): use conditional access to document

Slashgear:
## Why

All React component should be _Universal_ and should not consider that `document` nor `window` or any DOM related global object are defined in order to work in any react application context.

## How

- Handle conditional access to cookie from document in order to make the component render nicely on the server side.


philibea (approved):

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
```diff
@@ -72,7 +73,7 @@ export const CookieConsentProvider = ({
 }>) => {
   const [needConsent, setNeedsConsent] = useState(false)
   const [cookies, setCookies] = useState<Record<string, string>>(
-    cookie.parse(document.cookie),
+    IS_CLIENT ? cookie.parse(document.cookie) : {},
```
iManu, line 76: question: ça ne pose pas de souci avec le type Record quand le retour est l'objet vide du ternaire ?

iManu (approved):

packages/cookie-consent/src/CookieConsentProvider/CookieConsentProvider.tsx
Slashgear: Non, justement, un objet vide correspond à un record vide

PR #2053, opened 2024-06-20
fix(use-storage): using with SSR

gmonte:
I am using Nextjs 14 and I am receiving some errors on build time because this package are using `window` directly when getting the storage. I created the `getStorage()` function that use `canUseDOM` to check if we can access `window` and it returns the corresponding storage.

Now I am able to use it with Nextjs.


philibea:
Hello @gmonte, first thanks for your contribution.

it's will be review after 2 things:
 - add a [changeset ](https://github.com/changesets/changesets)that describe your change
 - add some test to keep our code coverage under the right threshold

packages/use-storage/src/index.ts
```diff
@@ -24,6 +24,34 @@ const canUseDOM =
     typeof window.sessionStorage !== 'undefined'
   )
 
+const createMemoryStorage = (): Storage => {
```
philibea, line 27: Missing unit test on this part

chambo-e:
Hello @gmonte 👋 

Thanks for your contrubution. I'm closing this, feel free to open an issue if you'd like to reopen it

PR #2112, opened 2024-07-30
feat: add type to locales and remove automatic mapping

alexandre-combemorel:


packages/use-i18n/src/__tests__/usei18n.test.tsx
```diff
@@ -20,7 +22,8 @@ type NamespaceLocale = {
 
 type OnTranslateError = ComponentProps<typeof I18n>['onTranslateError']
 
-const defaultSupportedLocales = ['en', 'fr', 'es']
+const isLocaleSupportedDefault = (locale: string): locale is Locales =>
```
philibea, line 25:
```suggestion
const isDefaultSettingsLocalesSupported = (locale: string): locale is Locales =>
```

sound better ?

packages/use-i18n/src/usei18n.tsx
```diff
@@ -93,8 +84,8 @@ const getCurrentLocale = ({
   return defaultLocale
 }
 
-type Context<Locale extends BaseLocale> = {
-  currentLocale: string
+type Context<LocaleParam extends BaseLocale, T extends string> = {
```
philibea, line 87: What is T ?

packages/use-i18n/src/__tests__/usei18n.test.tsx
alexandre-combemorel: Agree, I will just remove settings in the name, as we only check for the list of locales and not the whole settings list of locale

packages/use-i18n/src/usei18n.tsx
alexandre-combemorel: T is the union list of locale types supported. I will rename it

johnrazeur (approved):

packages/use-i18n/src/usei18n.tsx
```diff
@@ -23,53 +23,54 @@ import type { ReactParamsObject, ScopedTranslateFn, TranslateFn } from './types'
...
-        t: (str: 'You must pass a generic argument to useI18n()') => void
-        namespaceTranslation: (
-          str: 'You must pass a generic argument to useI18n()',
-        ) => void
-      }
-    : Context<Locale>
+type RequiredGenericContext<
+  LocaleParam extends BaseLocale,
+  LocalSupportedType extends string,
+> = keyof string extends never
```
philibea, line 34:
```suggestion
> = keyof LocaleParam extends never
```

Keyof string extends never will be always false ? We try to check the generic here

philibea (approved):

packages/use-i18n/src/usei18n.tsx
alexandre-combemorel: indeed ! good catch

Issue #1706, opened 2023-12-06
Fragile detection of the Renovate branch


moltar:
https://github.com/scaleway/scaleway-lib/blob/e2fe4046e01433273eb4356a508820b8bd9117c2/packages/changesets-renovate/src/index.ts#L98-L102

This branch prefix is configurable and may differ.

PR #2133, opened 2024-08-07
feat(changesets-renovate): add option to skip branch check

ImOverlord:
Hello,

I've added an option to disable the verification that it's running in a renovate branch `SKIP_BRANCH_CHECK`. This helps with projects that may have a custom branch name for renovate (#1706) or if you are using GitLab CI where you are in a detached HEAD state.

### DIFF
--- a/packages/changesets-renovate/src/index.ts
+++ b/packages/changesets-renovate/src/index.ts
@@ -94,11 +94,12 @@ async function getBumps(files: string[]): Promise<Map<string, string>> {
 
 export async function run() {
   const branch = await simpleGit().branch()
+  const branchPrefix = process.env['BRANCH_PREFIX'] ?? 'renovate/'
 
   console.log('Detected branch:', branch)
 
   if (
-    !branch.current.startsWith('renovate/') &&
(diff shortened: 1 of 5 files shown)


packages/changesets-renovate/src/index.ts
```diff
@@ -97,7 +97,10 @@ export async function run() {
 
   console.log('Detected branch:', branch)
 
-  if (!branch.current.startsWith('renovate/')) {
+  if (
+    !branch.current.startsWith('renovate/') &&
+    !process.env['SKIP_BRANCH_CHECK']
```
philibea, line 102: Maybe the idea behind https://github.com/scaleway/scaleway-lib/issues/1706 was to configure the name of the branch ? 

```ts
branch.current.startsWith('renovate/')
```
 per 
```
ranch.current.startsWith(BRANCH_NAME)
```

We can keep the SKIP_BRANCH_CHECK but it's will be more response to the issue your are tracking.

ImOverlord: Good point. I've added an option to have a custom branch name.

philibea (approved):

PR #2282, opened 2024-11-06
feat: add useInfiniteDataLoader

DorianMaliszewski:
Implement useInfiniteDataLoader to handle Load More on our UI

### DIFF
--- a/packages/use-dataloader/src/types.ts
+++ b/packages/use-dataloader/src/types.ts
@@ -97,21 +97,16 @@ export type UseDataLoaderResult<ResultType, ErrorType> = {
 export type UseInfiniteDataLoaderConfig<
   ResultType,
   ErrorType extends Error,
-  ParamsType,
+  ParamsType extends Record<string, unknown>,
   ParamsKey extends keyof ParamsType,
 > = Omit<UseDataLoaderConfig<ResultType, ErrorType>, 'initialData'> & {
   /**
-   * Params will be forwarded to method
-   */
-  params: ParamsType
-  /**
-   * The key to change in params
-   */
-  pageKey: ParamsKey
-  /**
    * If return undefined it consider that there are no remaining page to load
    */
-  getNextPage: (result: ResultType) => ParamsType[ParamsKey]
+  getNextPage: (
+    lastRes: ResultType,
+    lastParams: ParamsType,
+  ) => ParamsType[ParamsKey]
   /**
    * The initial data if no one is present in the cache before the request
    */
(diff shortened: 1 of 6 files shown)


alexandre-combemorel (approved):
I don't have much time to fully deep dive into the code flow. But happy to test it with our use cases. LGTM

packages/use-dataloader/src/useInfiniteDataLoader.ts
```diff
@@ -0,0 +1,227 @@
...
+    requestRefs.current = []
+    forceRerender()
+  }, [baseKey, forceRerender])
+
+  useEffect(
+    () => () => {
+      requestRefs.current.forEach(request => request.removeObserver(() => null))
+    },
+    [],
+  )
```
johnrazeur, line 81: Is it executed ?

packages/use-dataloader/src/types.ts
```diff
@@ -16,78 +16,117 @@ export type NeedPollingType<ResultType> =
...
- */
-export type UsePaginatedDataLoaderMethodParams = {
-  page: number
-  perPage: number
+export type UseInfiniteDataLoaderConfig<
+  ResultType,
+  ErrorType extends Error,
+  ParamsType,
+  ParamsKey extends keyof ParamsType,
+> = Omit<UseDataLoaderConfig<ResultType, ErrorType>, 'initialData'> & {
```
johnrazeur, line 102: Should we remove polling ?

packages/use-dataloader/src/useInfiniteDataLoader.ts
```diff
@@ -0,0 +1,227 @@
...
+
+  const isSuccess = request.status === StatusEnum.SUCCESS
+
+  const isError = request.status === StatusEnum.ERROR
+
+  const isIdle = requestRefs.current.every(
+    req => req.status === StatusEnum.IDLE && !enabled,
+  )
+
+  const reload = useCallback(async () => {
```
johnrazeur, line 137: What happen when a page doesn't exists anymore ?

DorianMaliszewski: data will be null/undefined/empty array and be filtered out when user get the data from the hook. May be if it doesnt fit some edge cases I can cont how many requests we have and try to get each one sequentially then by using getNextPage the hook can try to fill the same number of pages but if it cant it will not make useless calls ? :thinking:

philibea (approved):

packages/use-dataloader/src/types.ts
DorianMaliszewski: We keep it when a data is in transient state

johnrazeur (approved):

PR #2545, opened 2025-04-10
feat(changesets-renovate): filter out private packages

silouanwright:
In a monorepo, people may choose to have some packages published, and keep other packages private.

Private packages (which are represented with a `private: true` key) ideally shouldn't be published. An example below:

```json
{
  "name": "@scope/package-name",
  "version": "0.0.0",
  "private": true,
  "description": "An internal package for providing a shared configuration to apps",
  "main": "./index.js",
}
```

[In the original GitHub Action from Backstage this is based off](https://github.com/backstage/backstage/blob/master/.github/workflows/sync_renovate-changesets.yml), the code is in fact filtering out private packages:

```typescript
// Parses package.json files and returns the package names
async function getPackagesNames(files) {
  const names = [];
  for (const file of files) {
    const data = JSON.parse(await fs.readFile(file, "utf8"));
    if (!data.private) { // ← It only grabs the package if it's not private.
      names.push(data.name);
    }
  }
  return names;
}
```

Would love to see that functionality represented in this package as well.

### DIFF
--- a/.changeset/clean-tigers-stalk.md
+++ b/.changeset/clean-tigers-stalk.md
@@ -1,0 +1,5 @@
+---
+"@scaleway/changesets-renovate": minor
+---
+
+filter out private packages from being published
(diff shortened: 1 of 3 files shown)


silouanwright:
👋

chambo-e:
Hello @silouanwright 

Sorry for the delay

We are going to handle private packages the same way changeset does

You can follow the feature here https://github.com/scaleway/scaleway-lib/pull/3288 which should be available very soon

Thanks for your contribution

PR #2646, opened 2025-06-18
chore: use `.toBeCallableWith()` in type tests

mrazauskas:
To draw attention, TSTyche 4 shipped with `.toBeCallableWith()` as well as `.not.toBeCallableWith()` matchers.

These can be used instead of `.toRaiseError()` to communicate the intent of the test much better. That also means there will be no red squiggles, because TypeScript will not detect any errors.

---

If there any errors would appear in the type tests (try removing some `.not`), TSTyche will be loud and clear about that:
 
Screenshot 2025-06-18 at 14 45 52

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v5.0.0
         with:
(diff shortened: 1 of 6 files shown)


mrazauskas:
@philibea Could you take a look, please? (;

PR #2728, opened 2025-07-29
feat(regex): update regex for backupkey

JulienSaguez:
update regex for backupkey


BABAK0T0 (approved):

packages/regex/src/__tests__/index.test.ts
```diff
@@ -521,7 +521,7 @@ describe('@regex', () => {
       [backupKeyTest, true],
       [digitsTest, false],
       [emailTest, false],
-      [octdigits, false],
+      [octdigits, true],
```
philibea, line 524: add 32 maybe ?

PR #3154, opened 2026-04-03
feat(vite+): migrate toolchain to vite+

philibea:


chambo-e:
I think we can close this as we migrated to tsdown :)

PR #3253, opened 2026-05-06
docs: add instructions to test a local package in a pnpm workspace

jsulpis:
I used pnpm overrides to test ultraviolet packages inside a pnpm workspace, I thought it could be useful to be documented here. I guess it works the same for scaleway-lib packages.


chambo-e (approved):
Thanks for that

This readme as a whole needs a bit of a cleanup, I'll do that asap 😄

PR #3257, opened 2026-05-07
build: replace vite with tsdown

chambo-e:
## Changes
- Packages are now ESM only 
- CLI packages are now bin only
- Node < 24 is dropped
- npmignore handling has been centralised

## Gains
- build is significantly faster (2ish seconds vs 12/13)
- [publint](https://publint.dev/) and [AreTheTypesWrong](https://arethetypeswrong.github.io/) are now baked in the build process.

## Output diff
Output is 99% iso. Only things that I saw: 
- `var` becomes `const` => Should have already been the case
- enums are now inlined => No impact

Screenshot 2026-05-13 at 15 29 41


etienne-scaleway:
**suggestion**: Use https://tsdown.dev/options/package-exports everywhere ?

packages/use-growthbook/__mocks__/@growthbook/growthbook-react.tsx
```diff
@@ -2,20 +2,20 @@ import type { ReactNode } from 'react'
 import { vi } from 'vitest'
 
 // Create mock functions
-export const init = vi.fn()
-export const getAttributes = vi.fn()
-export const setAttributes = vi.fn()
-export const loadFeatures = vi.fn()
+export const init: ReturnType<typeof vi.fn> = vi.fn()
```
alexandre-combemorel, line 5: Tu avais des erreurs de type explicite au build de la lib ?

chambo-e: Yup, unfortunately it has been here for a while, turbo output made it difficult to see them

packages/changesets-renovate/package.json
```diff
@@ -12,35 +12,25 @@
...
-      "default": "./dist/cli.js"
-    }
-  },
+  "module": "./dist/cli.mjs",
   "publishConfig": {
     "access": "public"
   },
   "scripts": {
-    "prebuild": "shx rm -rf dist",
+    "prepack": "cp ../../.npmignore .",
```
jsulpis: could it be a symbolic link maybe ?

packages/changesets-renovate/package.json
```diff
@@ -12,35 +12,25 @@
...
     "typecheck": "tsgo --noEmit",
-    "type:generate": "tsgo --declaration -p tsconfig.build.json",
-    "build": "vite build --config vite.config.ts && pnpm run type:generate",
-    "build:profile": "npx vite-bundle-visualizer -c vite.config.ts",
-    "test:unit": "vitest --run --config vite.config.ts",
+    "build": "tsdown --config tsdown.config.ts",
```
jsulpis: I guess tsdown will find the config automatically, why repeat it everywhere ? 

```suggestion
    "build": "tsdown",
```

packages/fuzzy-search/package.json
```diff
@@ -5,31 +5,27 @@
...
   "publishConfig": {
     "access": "public"
   },
   "scripts": {
-    "prebuild": "shx rm -rf dist",
+    "prepack": "cp ../../.npmignore .",
     "typecheck": "tsgo --noEmit",
-    "type:generate": "tsgo --declaration -p tsconfig.build.json",
-    "build": "vite build --config vite.config.ts && pnpm run type:generate",
-    "build:profile": "npx vite-bundle-visualizer -c vite.config.ts",
```
jsulpis, line 31: no bundle visualizer anymore ?

chambo-e:
> **suggestion**: Use https://tsdown.dev/options/package-exports everywhere ?

@etienne-scaleway good catch, added :)

Only disabled on CLI packages because there is no need to have outputs outside of the cli

packages/use-analytics/package.json
```diff
@@ -21,36 +21,19 @@
...
-    "./cookies-consent": {
-      "types": "./dist/cookies-consent/index.d.ts",
-      "require": "./dist/cookies-consent/index.cjs",
-      "import": "./dist/cookies-consent/index.js",
-      "default": "./dist/cookies-consent/index.js"
-    }
+    ".": "./dist/index.js",
+    "./analytics": "./dist/analytics/index.js",
+    "./cookies-consent": "./dist/cookies-consent/index.js",
+    "./package.json": "./package.json"
```
jsulpis: no types ?

LICENSE.md
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Scaleway
+Copyright (c) 2021-2026 Scaleway
```
jsulpis: It seems that this can be the year of the creation of the asset, not necessarily a range (which will become outdated quickly)
https://opensource.stackexchange.com/a/1523

tsdown.config.ts
```diff
@@ -0,0 +1,16 @@
...
+    tsgo: true,
+  },
+  target: 'baseline-widely-available',
+  platform: 'neutral',
+  entry: 'src/index.ts',
+  attw: {
+    profile: 'esm-only',
+  },
+  publint: true,
+  unbundle: true,
```
jsulpis, line 14: do we want this behavior by default ? It generates more files so I think it can lead to performance issues especially with barrel files (but I don't know the packages in this repo, maybe it's no big deal)

packages/fuzzy-search/package.json
chambo-e: Not yet supported by tsdown https://github.com/rolldown/rolldown/issues/6425

Also I'm not sure anyone was using it here, the packages are quite straightforward :)

tsdown.config.ts
chambo-e: Packages here are quite simple, I'd like to do the least bundling and transformation 

Also, barrel performance should not really be a concern here, outside of root file exposing lib interface I don't think we have them :)

packages/changesets-renovate/package.json
chambo-e: Good idea :)

packages/use-analytics/package.json
chambo-e: Fixed, thx

jsulpis (approved):
simpler, better, faster, stronger 🤖🎵

PR #3266, opened 2026-05-07
feat(oxfmt-config): add mergeConfig function

chambo-e:


.changeset/odd-peaches-read.md
```diff
@@ -0,0 +1,5 @@
+---
+"@scaleway/oxfmt-config": minor
+---
+
+add mergeConfig function
```
jsulpis: you should mention that it adds a default ignorePatterns too

packages/oxfmt-config/README.md
```diff
@@ -13,13 +13,23 @@ npm install --save-dev @scaleway/oxfmt-config
...
+```ts
+import { defineConfig } from 'oxfmt'
+import config, { mergeConfig } from '@scaleway/oxfmt-config'
+
+export default mergeConfig(
+  config,
+  defineConfig({
+    ignorePatterns: ['custom_path/**'],
+  }),
+)
```
jsulpis, line 32: I'm not sure it's needed to add a dependency for that, if one day someone wants to override or extend the ignorePattern they can just spread it like before, no ?

```suggestion
export default defineConfig({
  ...config,
  ignorePatterns: [
    ...config.ignorePatterns,
    "custom_path/**"
  ]
})
```

chambo-e: I don't really want to handle all edge cases with each config vars 😄 ignorePatterns is only one of many that we could extend

https://oxc.rs/docs/guide/usage/formatter/config-file-reference.html

jsulpis (approved):

packages/oxfmt-config/README.md
jsulpis: alright 👍

PR #3365, opened 2026-05-25
feat(phonenumber): add awesome-phonenumber lib

philibea:


packages/phonenumber/.npmignore
```diff
@@ -0,0 +1,7 @@
+**/__tests__/**
```
chambo-e: Should be a link :)

packages/react-router/.npmignore
```diff
@@ -0,0 +1,7 @@
+**/__tests__/**
```
chambo-e: Should also be a link :)

chambo-e (approved):

chambo-e (approved):

chambo-e (approved):

PR #3466, opened 2026-06-24
fix: use infinite dataloader stale page

dmzski:
## Issue

- Actually when you fetch multiple base with a hook then change a subcomponent and you baseParams the query will fetch the first page but not clean the nextPageToken so I will duplicate the result or lead to an error

## What did I do
- Ensure the baseQueryKey still the same or reset page and nextPageRef


philibea:
@dmzski can you add a test that show the issue ?

chambo-e (approved):

PR #3481, opened 2026-06-29
chore: add auth scw package

alexandre-combemorel:


packages/auth-scw/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "description": "Auth Scaleway",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/scaleway/scaleway-lib.git",
+    "directory": "packages/auth-scw"
+  },
+  "type": "module",
+  "exports": {
+    ".": "./dist/index.mjs",
```
philibea: mjs ?

packages/auth-scw/src/types/account/index.ts
```diff
@@ -0,0 +1,410 @@
+// eslint-disable-next-line no-abusive-eslint-disable
+/* oxlint-disable */
+import type { UnauthenticatedUserAPI } from './api'
```
philibea: can we use only what we need ?

packages/auth-scw/src/useAuthScw/createClient.ts
```diff
@@ -0,0 +1,72 @@
...
+  SESSION_HEADER_KEY,
+} from '@scaleway/sdk-client'
+import type { ConfigurationClientMethods, CreateTemporaryClientType, ConfigAuthProvider } from '../types'
+
+type ClientSingletonType = {
+  setAPIsAndSettings: (params: ConfigAuthProvider) => void
+  createClient: (params: CreateTemporaryClientType) => ConfigurationClientMethods
+} & Partial<ConfigAuthProvider>
+
+export const clientSingleton: ClientSingletonType = {
```
philibea, line 15: this is not really a singleton but a simily class here ? 

we only use clientSingleton.createClient ? the singleton is only the settings ?

packages/auth-scw/src/types/account/index.ts
alexandre-combemorel: ye je vais clean up

packages/auth-scw/src/useAuthScw/createClient.ts
alexandre-combemorel: This is to identify the purpose of it, it is not a real singleton but it is use as is.

packages/auth-scw/mocks/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  mockInitiateOidcLogin,
+  mockStartWebAuthnAuthentication,
+  mockInitiatePasswordlessLogin,
+  mockLoginOrganizations,
+  mockLoginSession,
+  mockLogout,
+  mockRenewToken,
+  useAuthScw,
+  AuthStoreManager,
+  useGetDediboxRedirectionUrlWithToken,
```
philibea: not use

packages/auth-scw/mocks/index.ts
```diff
@@ -0,0 +1,144 @@
...
+
+const useAuthScw = vi.fn<() => AuthScwContextType>(() => defaultMockUseAuthScw)
+
+const AuthScwProvider = ({ children }: { children: ReactNode }) => children
+
+const AuthStoreManager = {
+  getJwt: vi.fn((): JWT | null => MOCK_ENCODED_JWT_COOKIE.jwt),
+}
+
+const useGetDediboxRedirectionUrlWithToken = vi.fn(() => vi.fn((url: string) => `${url}`))
```
chambo-e: This hook seems to be missing

chambo-e (approved):

chambo-e (changes requested):
Sorry bad manip

packages/auth-scw/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "devDependencies": {
+    "@scaleway/sdk-client": "catalog:",
+    "@types/react": "catalog:",
+    "cookie": "catalog:",
+    "date-fns": "catalog:",
+    "react": "catalog:",
+    "zod": "catalog:"
+  },
+  "peerDependencies": {
+    "@scaleway/sdk-client": "catalog:",
```
philibea, line 44: on peerDeps we need more flexibility, this will enforce a pinned version here

packages/auth-scw/src/__tests__/isExpired.test.ts
```diff
@@ -0,0 +1,30 @@
+import { describe, expect, it } from 'vitest'
```
chambo-e, line 1: To be consistent with the others tests this should be put in a __tests__ folder ?

packages/auth-scw/package.json
```diff
@@ -0,0 +1,37 @@
+{
```
chambo-e, line 1: Missing homepage, bugs, types, publishConfig and maybe sideEffects ?

packages/auth-scw/package.json
alexandre-combemorel: let's keep it in catalog and if required we can fix it in the package. We leverage the use of catalog efficiently this way

philibea (approved):

packages/auth-scw/package.json
```diff
@@ -0,0 +1,55 @@
...
+    ".": "./dist/index.js",
+    "./package.json": "./package.json"
+  },
+  "publishConfig": {
+    "exports": {
+      ".": "./dist/index.js",
+      "./package.json": "./package.json"
+    },
+    "access": "public"
+  },
```
chambo-e, line 27:
```suggestion
  "publishConfig": {
    "access": "public"
  },

```

packages/auth-scw/package.json
```diff
@@ -0,0 +1,55 @@
+{
```
chambo-e, line 1: .npmignore file is missing :)

packages/auth-scw/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "@scaleway/tsconfig",
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "ESNext"
+  },
```
chambo-e: Why ? 🤔

packages/auth-scw/vitest.config.ts
```diff
@@ -0,0 +1,11 @@
+import { mergeConfig } from 'vitest/config'
+import { defaultConfig } from '../../vitest.config'
+
+export default mergeConfig(defaultConfig, {
+  test: {
+    // We use happy-dom here because jsdom does not handle navigation
```
chambo-e, line 6: We can use jsdom here no ? 🤔

pnpm-workspace.yaml
```diff
@@ -2,26 +2,26 @@ packages:
...
-  '@testing-library/jest-dom': 6.9.1
-  '@testing-library/react': 16.3.2
-  '@types/node': 25.9.4
-  '@types/picomatch': 4.0.3
-  '@types/react': 19.2.17
-  '@types/react-dom': 19.2.3
-  '@types/react-router-dom': 5.3.3
-  '@vitejs/plugin-react': 6.0.3
-  '@vitest/coverage-istanbul': 4.1.9
+  "@arethetypeswrong/core": 0.18.4
```
chambo-e: I think format is wrong here, it should be single quote :)

pnpm-workspace.yaml
```diff
@@ -69,12 +70,16 @@ allowBuilds:
...
 
 supportedArchitectures:
-  cpu: [ x64, arm64 ]
-  os: [ darwin, current ]
+  cpu: [x64, arm64]
+  os: [darwin, current]
+
+minimumReleaseAgeExclude:
+  - "@scaleway/sdk-iam@2.9.4"
+  - "@scaleway/sdk-std@2.4.3"
```
chambo-e:
```suggestion
  - "@scaleway/sdk-iam"
  - "@scaleway/sdk-std"
```

packages/auth-scw/mocks/index.ts
```diff
@@ -0,0 +1,61 @@
+import type { ReactNode } from 'react'
```
chambo-e: It seems that only MOCK_AUDIENCE_ID and MOCK_ENCODED_JWT_COOKIE are used from this file ? Can we remove the rest ?

packages/auth-scw/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "name": "@scaleway/auth-scw",
+  "version": "1.0.0",
```
chambo-e: Version should be < 1.0 and bumped with a major changeset

chambo-e (changes requested):
Also nitpick: shouldn't this package be named `@scaleway/use-auth` as the hook is the main export ?

packages/auth-scw/tsconfig.json
alexandre-combemorel: Il s'avérait que j'avais des soucis de typecheck en local, mais c'était du a mon lsp tsgo qui partait en sucette, l'ajout de la target avait fixé le soucis, mais le problème était plus que le lsp était planté.
Je supprime

packages/auth-scw/vitest.config.ts
alexandre-combemorel: nop, un des test utilise le pkg history et plante avec jsdom. Potentiellement refacto possible mais ca sera fait à part

packages/auth-scw/mocks/index.ts
alexandre-combemorel: yes, I was wondering if it could be useful to have those mock exported and be reused in a repo for testing where it is integrated, but I agree it can be done directly into a repo where it is used

chambo-e (approved):

PR #3505, opened 2026-07-09
fix: auth-scw deps and package.json

alexandre-combemorel:
- switch from peerDeps to dependecies
- remove sideEffect false, which is not true

### DIFF
--- a/packages/auth-scw/package.json
+++ b/packages/auth-scw/package.json
@@ -44,9 +44,7 @@
   },
   "peerDependencies": {
     "@scaleway/sdk-client": "catalog:",
+    "@scaleway/sdk-iam": "catalog:",
     "react": "catalog:"
-  },
-  "inlinedDependencies": {
-    "@scaleway/sdk-iam": "2.9.4"
   }
 }
--- a/packages/auth-scw/package.json
+++ b/packages/auth-scw/package.json
@@ -16,7 +16,7 @@
     "dist/*"
   ],
   "type": "module",
-  "sideEffects": false,
+  "sideEffects": true,
   "types": "./dist/index.d.ts",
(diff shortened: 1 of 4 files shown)


packages/auth-scw/package.json
```diff
@@ -31,21 +31,22 @@
...
+    "react": "catalog:"
   },
   "peerDependencies": {
     "@scaleway/sdk-client": "catalog:",
-    "@scaleway/sdk-iam": "catalog:",
-    "cookie": "catalog:",
-    "date-fns": "catalog:",
-    "react": "catalog:",
-    "zod": "catalog:"
+    "react": "catalog:"
```
chambo-e: Should be a range instead of a fixed version no ?

pnpm-workspace.yaml
```diff
@@ -2,28 +2,28 @@ packages:
...
-  "@testing-library/jest-dom": 6.9.1
-  "@testing-library/react": 16.3.2
-  "@testing-library/user-event": 14.6.1
-  "@types/node": 25.9.4
-  "@types/picomatch": 4.0.3
-  "@types/react-dom": 19.2.3
-  "@types/react": 19.2.17
-  "@vitejs/plugin-react": 6.0.3
-  "@vitest/coverage-istanbul": 4.1.9
+  '@arethetypeswrong/core': 0.18.4
```
chambo-e: bad fmt :)

alexandre-combemorel: that's the problem, it seems that yaml files do not all follow the same format. i can format it the way fmt does, but it's in conflict with others generated yaml

chambo-e (approved):

PR #3634, opened 2026-08-17
feat(oxlint-config): new global configs

chambo-e:


alexandre-combemorel (approved):

packages/oxlint-config/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import oxc from './plugins/oxc'
+import reactConfig from './plugins/react'
+import typescript from './plugins/typescript'
+import unicorn from './plugins/unicorn'
+import vitestConfig from './plugins/vitest'
+
+export const base = defineConfig({
+  options: {
+    typeAware: true,
+    typeCheck: true,
```
philibea: do we want to activate typecheck by default ?

packages/oxlint-config/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    reportUnusedDisableDirectives: 'error',
```
philibea, line 14: can we set to warn instead ?

packages/oxlint-config/src/index.ts
chambo-e: I think yes no ? This is one of the big advantages of oxlint and if we disable it half the typescript rules we've set to error would be disabled

packages/oxlint-config/src/index.ts
chambo-e: Why tho ? If a directive is unused it should be removed no ?

philibea (approved):

PR #3660, opened 2026-08-27
feat(utility-types): add types

philibea:


packages/types/.npmignore
```diff
@@ -0,0 +1,9 @@
+**/__tests__/**
```
chambo-e: Needs to be a symlink

packages/types/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@scaleway/utility-types",
```
chambo-e: Just no ? 

```suggestion
  "name": "@scaleway/types",
```

packages/types/tsdown.config.ts
```diff
@@ -0,0 +1,13 @@
...
+export default defineConfig({
+  ...defaultConfig,
+  copy: [
+    {
+      from: 'src/global.d.ts',
+      to: 'dist',
+      rename: 'global.d.ts',
+    },
+  ],
+})
```
chambo-e: Why not use index.ts ? 

Types should be properly exported

chambo-e (approved):

alexandre-combemorel (approved):

PR #3681, opened 2026-08-31
fix: properly ignore all non git tracked

chambo-e:


packages/changesets-renovate/src/globWithGitignore.ts
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env node
```
chambo-e, line 1: No need for a shebang here

philibea (approved):

PR #3798, opened 2026-09-05
fix(oxlint): fix vitest/prefer-expect-type-of violations

### LINKED ISSUE #3776: [oxlint] Fix `vitest/prefer-expect-type-of` violations
This issue is part of the [Oxlint v1](https://github.com/scaleway/scaleway-lib/milestone/2) effort to clean up the warning rules in `oxlint.config.ts`.

## Rule

`vitest/prefer-expect-type-of`

**Documentation:** https://oxc.rs/docs/guide/usage/linter/rules/vitest/prefer-expect-type-of

## Context

This rule is currently set to `warn` in the `overrides` block (test files) of [`oxlint.config.ts`](https://github.com/scaleway/scaleway-lib/blob/main/oxlint.config.ts). The comment at the top of the `rules` block states that all custom rules below should be removed once the violations are fixed.

## Goal

1. Fix all violations of `vitest/prefer-expect-type-of` across the codebase.
2. Remove the `vitest/prefer-expect-type-of` line from `oxlint.config.ts`.
3. Ensure `oxlint` and `typecheck` pass.

## Workflow notes

- Do **not** open a PR if another PR for this rule is already open.
- Address any human review comments on an open PR with a new commit on the same branch.

philibea:
Closes #3776

Fixed all violations of `vitest/prefer-expect-type-of` and removed the rule from `oxlint.config.ts`.

Part of [Oxlint v1](https://github.com/scaleway/scaleway-lib/milestone/2).

### DIFF
--- a/packages/scouter/src/__tests__/BrowserRouter.test.tsx
+++ b/packages/scouter/src/__tests__/BrowserRouter.test.tsx
@@ -1,4 +1,4 @@
-// oxlint-disable typescript/no-unsafe-assignment typescript/no-unsafe-type-assertion typescript/no-explicit-any vitest/no-conditional-in-test
+// oxlint-disable typescript/no-unsafe-assignment typescript/no-unsafe-type-assertion typescript/no-explicit-any vitest/no-conditional-in-test typescript/unbound-method
 import { render, screen } from '@testing-library/react'
 import { describe, expect, it } from 'vitest'
 import type { History, Location, Match } from '../index'
@@ -36,8 +36,8 @@ describe('component BrowserRouter', () => {
     )
 
     expect(history).toBeDefined()
-    expect(typeof history.push).toBe('function')
-    expect(typeof history.replace).toBe('function')
+    expect(history.push).toBeTypeOf('function')
+    expect(history.replace).toBeTypeOf('function')
   })
 
   it('history is re-created for each BrowserRouter', () => {
@@ -108,6 +108,6 @@ describe('component BrowserRouter', () => {
     )
 
     expect(location!).toBeDefined()
-    expect(typeof location!.pathname).toBe('string')
+    expect(location!.pathname).toBeTypeOf('string')
   })
 })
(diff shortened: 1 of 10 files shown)


packages/scouter/src/__tests__/BrowserRouter.test.tsx
```diff
@@ -1,4 +1,4 @@
-// oxlint-disable typescript/no-unsafe-assignment typescript/no-unsafe-type-assertion typescript/no-explicit-any vitest/no-conditional-in-test
+// oxlint-disable typescript/no-unsafe-assignment typescript/no-unsafe-type-assertion typescript/no-explicit-any vitest/no-conditional-in-test typescript/unbound-method
```
chambo-e: 💩

chambo-e (changes requested):

packages/scouter/src/__tests__/BrowserRouter.test.tsx
philibea: Addressed in d1e88d58. Removed `typescript/unbound-method` from all file-level disable comments.

For method references (`history.push`, `route.link`, `result.current.push`, etc.), reverted to `expect(typeof X).toBe('function')` with inline `oxlint-disable-next-line vitest/prefer-expect-type-of`. This is a legitimate conflict: `toBeTypeOf` passes the method reference to `expect()`, which triggers `unbound-method`. Using `typeof` is the safe way to check a method's type without an unbound reference.

Non-method properties (strings, objects, plain functions) keep `toBeTypeOf` as the clean fix — they don't trigger `unbound-method`.

packages/scouter/src/__tests__/BrowserRouter.test.tsx
```diff
@@ -36,7 +36,9 @@ describe('component BrowserRouter', () => {
     )
 
     expect(history).toBeDefined()
+    // oxlint-disable-next-line vitest/prefer-expect-type-of -- method ref triggers unbound-method
```
chambo-e, line 39: Not quite better 😄

PR #3829, opened 2026-09-09
fix(oxlint-config): typescript/strict-boolean-expressions is now more flexible and only checks for bool/any/enums

chambo-e:


packages/oxlint-config/src/plugins/typescript.ts
```diff
@@ -421,7 +421,19 @@ export default defineConfig({
     /**
      * @see https://oxc.rs/docs/guide/usage/linter/rules/typescript/strict-boolean-expressions
      */
-    'typescript/strict-boolean-expressions': 'error',
+    'typescript/strict-boolean-expressions': [
+      'error',
+      {
+        allowAny: false,
+        allowNullableBoolean: false,
+        allowNullableEnum: false,
```
philibea, line 429: true

packages/oxlint-config/src/plugins/typescript.ts
```diff
@@ -421,7 +421,19 @@ export default defineConfig({
...
+        allowNullableNumber: true,
+        allowNullableObject: true,
+        allowNullableString: true,
+        allowNumber: true,
```
philibea, line 433: false

philibea (approved):

